PR #21, opened 2022-04-14
Allow full range of tag characters

jryans:
Some clients use tag characters beyond the basic word set (`\w`). This attempts to match the actual IMAP syntax for tags.

### DIFF
--- a/emailproxy.py
+++ b/emailproxy.py
@@ -79,9 +79,9 @@ AUTHENTICATION_TIMEOUT = 600
 
 TOKEN_EXPIRY_MARGIN = 600  # seconds before its expiry to refresh the OAuth 2.0 token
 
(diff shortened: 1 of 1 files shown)


jryans:
For comparison, here's the relevant bits from the [IMAP syntax](https://www.ietf.org/rfc/rfc9051.html#name-formal-syntax):

```
tag             = 1*<any ASTRING-CHAR except "+">
ASTRING-CHAR    = ATOM-CHAR / resp-specials
ATOM-CHAR       = <any CHAR except atom-specials>
atom-specials   = "(" / ")" / "{" / SP / CTL / list-wildcards /
                  quoted-specials / resp-specials
list-wildcards  = "%" / "*"
quoted-specials = DQUOTE / "\"
resp-specials   = "]"
DQUOTE          = %x22
SP              = %x20
CTL             = %x00-1F / %x7F
CHAR            = %x01-7F
```

jryans:
@simonrob Thanks for the review and creating this project! 😄 I have updated this as you suggested.

PR #22, opened 2022-04-15
Permit only valid IMAP tag characters rather than rejecting invalid

simonrob:
@jryans I realised that even with the tweak I suggested to the new IMAP tag matcher, it still allows invalid tags (e.g., anything beyond the basic 128 ASCII characters). That is still better than what was used before, but I think it's worth fully matching the formal specification, so the version in this pull request takes an allowlist-based approach instead.

Could you test this change with your client that uses non-alphanumeric tags?


jryans:
@simonrob Ah, good catch! I can confirm this version also works well for me. 😄

simonrob:
Great - will merge this change too. Thanks again for the original fix.

PR #88, opened 2022-10-31
Set client credential flow access token to 'None'

lakelse:
When the client credential access token expiry is within the expiry margin, set it to None.

### DIFF
--- a/emailproxy.py
+++ b/emailproxy.py
@@ -411,8 +411,8 @@ class OAuth2Helper:
                                        OAuth2Helper.encrypt(fernet, response['refresh_token']))
                         AppConfig.save()
 
-                    elif access_token_expiry <= current_time:
-                        access_token = None  # avoid trying invalid tokens
+                    else:
+                        access_token = None  # avoid trying invalid (or soon to be) tokens
                 else:
                     access_token = OAuth2Helper.decrypt(fernet, access_token)
 


simonrob:
Can you explain why this would be needed?

This condition is only currently encountered if there is no refresh token present, in which case either we automatically get a new token regardless (CCG / ROPCG flow) or will need to prompt the user in `TOKEN_EXPIRY_MARGIN` seconds regardless. What is the benefit from invalidating the session early and prompting the user here?

(I can understand why it might be good to retain the existing token _and_ prompt the user early, but that is not what this change does.)

lakelse:
Hey @simonrob, first of all, great work on this project!

So, I can only speak to my experience using the client credentials grant flow.  When the access token falls into the condition evaluating the expiry, there's no refresh token of course...  Then on to the 'elif' checking if the token expiry is less than the current_time...  If it's not, the token is never decrypted.  Perhaps instead it should be?  Forgive my lack of understanding here but 
I do know there's bug with the client credentials flow if the access token is within the expiry margin, it'll result in an error because it tries to use the encrypted access token.

simonrob:
> I do know there's bug with the client credentials flow if the access token is within the expiry margin, it'll result in an error because it tries to use the encrypted access token.

This is the key part – thanks for pointing this out. I would rather keep the token to use while it is still valid, but happy to accept a change that fixes use of the token within the expiry period (e.g., add an additional `else` block to decrypt)

simonrob:
Actually, scratch that - the whole point of the expiry period is to refresh the token. I'll merge this request.

lakelse:
Thanks, Simon!  It's a pleasure to make a contribution. I'm sure you've saved people untold hours.

PR #100, opened 2022-11-24
Docker image

interone-ms:
Hi,

I have created a Dockerfile that uses your recent addition of `pyinstaller` to produce a minmal-sized Docker image (~20 MB instead of ~140 MB).

Unfortunately I'm not experienced with GitHub Actions, but it should be possible for you to set up an automated Docker build of this image and a publish on Docker Hub.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,16 @@
+FROM python:alpine as builder
+
+WORKDIR /email-oauth2-proxy
+
+COPY requirements-no-gui.txt requirements-no-gui.txt
+COPY emailproxy.py emailproxy.py
+
+RUN	apk add curl busybox-extras binutils && \
+	python -m pip install -r requirements-no-gui.txt && \
+	pip install -U pyinstaller && \
+	pyinstaller --onefile emailproxy.py
+
+FROM alpine:3.16 as app-base
+COPY --from=builder /email-oauth2-proxy/dist/emailproxy /usr/local/bin/emailproxy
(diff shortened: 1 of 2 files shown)


simonrob:
Thanks for the work on this, and for the pull request.

It's great to see a more efficient Docker setup for the proxy, but I'd prefer not to merge configurations for specific external container setups to avoid taking on the ongoing maintenance burden.

If you plan to keep it updated I am happy to link to your fork from the [related projects](https://github.com/simonrob/email-oauth2-proxy#related-projects-and-alternatives) section of the readme.

ski7777:
Hey @simonrob,

I'd really like to continue on this approach. What do you want to merge this PR?
- autobuild with github workflow
- setting `local_address = 0.0.0.0` automatically for example by introducing a new cli argument

Is this enough or are there any other blocking points?

moriahmorgan:
@ski7777  this has already been accomplished as seen in the readme. 

See this repo:

https://github.com/blacktirion/email-oauth2-proxy-docker

ski7777:
yes, but this is a different project and only builds manually. My plan is to build on each push or tag . Additionally we could implement an automatic push to pypi

simonrob:
I'm reluctant to take on the additional burden of maintaining a Docker configuration, since I don't use this myself.

The [repository](https://github.com/blacktirion/email-oauth2-proxy-docker) maintained by @blacktirion is, as explained above, essentially a ready-made version of this, and that is what I'd recommend at the moment.

The proxy is not updated especially often because it is essentially feature-complete, so I really wouldn't worry about not having the very latest commit.

PR #114, opened 2022-12-19
Add support for storing OAuth2.0 tokens in AWS Secrets Manager

### LINKED ISSUE #33: Storing OAuth token in secrets manager?
I'm opening this issue to inquire whether you would be open to merging a PR that implements an option to store the OAuth token in a Secrets Manager (e.g. AWS Secrets Manager, Github Secrets or Terraform Cloud Workspace Variables) instead of the local config. This would facilitate managing OAuth authorization and token creation from a local computer, while making the token seamlessly available to a remote server running `email-oauth2-proxy`. It would also facilitate persisting the token if the remote server needs to be destroyed and recreated.

As mentioned in a prior issue, I've been working on an infrastructure-as-code project which uses Terraform to automatically configure an AWS server running `email-oauth2-proxy`. I've just made a first version of my work public at: [michaelstepner/email-oauth2-proxy-aws](https://github.com/michaelstepner/email-oauth2-proxy-aws).

As you might anticipate, the biggest limitation of my automated server configuration is authenticating with the email provider to obtain OAuth tokens. Thanks to your prior work, after creating the lightweight server I'm able to authenticate manually via SSH using `--no-gui` and `--local-server-auth`. But I would love to configure a more elegant solution someday.

I wouldn't ask you to implement this yourself, of course, unless you were interested in doing so.

## Implementation

I considered a command-line argument, but I'm currently envisioning a per-account configuration option like `token_store` below:

```
[your.office365.address@example.com]
permission_url = https://login.microsoftonline.com/common/oauth2/v2.0/authorize
token_url = https://login.microsoftonline.com/common/oauth2/v2.0/token
oauth2_scope = https://outlook.office365.com/IMAP.AccessAsUser.All https://outlook.office365.com/SMTP.Send offline_access
redirect_uri = http://localhost
client_id = *** your client id here ***
client_secret = *** your client secret here ***
token_store = aws_secrets_manager
```

If that configuration option were left unspecified, it would default to `local` and the behaviour would follow the status quo, storing the token in the config file. This would preserve backward compatibility, and maintain simplicity for the majority of users who are using this software as a local client in their system tray.

Additional configuration required for the secrets manager could either be via environment variables (e.g. credentials for access to the secrets manager) or the configuration file.

michaelstepner:
Closes #33, which features a substantive discussion about the design of this feature.

# Objective

Provides an option to store tokens remotely in AWS Secrets Manager. By storing the authorization tokens remotely, this facilitates the administration of the proxy in a server-based configuration:

- individual servers can be replaced without reauthorizing: a new server can fetch the remote tokens and remain authenticated as long as the tokens haven't expired
- authenticating on a local computer can provide tokens to a headless server

# Implementation

This implementation endeavours to minimize the complexity added to the proxy:

1. It adds an optional account-level configuration parameter, `aws_secret`:

```
[your.office365.address@example.com]
permission_url = https://login.microsoftonline.com/common/oauth2/v2.0/authorize
token_url = https://login.microsoftonline.com/common/oauth2/v2.0/token
oauth2_scope = https://outlook.office365.com/IMAP.AccessAsUser.All https://outlook.office365.com/SMTP.Send offline_access
redirect_uri = http://localhost
client_id = *** your client id here ***
client_secret = *** your client secret here ***
aws_secret = *** ARN or name of secret in AWS Secrets Manager ***
```

2. The only dependency is `boto3`, and it is only loaded if one or more accounts have the `aws_secret` option specified in their config. This new dependency is documented in **requirements-aws-secrets.txt**

3. It adds new methods to the `AppConfig` class to handle the option of loading/saving OAuth 2.0 tokens from the remote AWS Secrets Manager.

    - These methods are only called from `AppConfig._load()` and `AppConfig.save()` if an account has the `aws_secret` option specified, which minimizes the risk that this new feature affects existing functionality.
    - These new methods are all named with the prefix `aws_secrets`.

4. If the proxy fails to access the AWS Secrets Manager, it logs an error but continues functioning normally: without fetching the remote secret, or without storing the secret remotely.

# Manual Testing

I've tested this new feature quite carefully using the Terraform config in [PR #1 of michaelstepner/email-oauth2-proxy-aws](https://github.com/michaelstepner/email-oauth2-proxy-aws/pull/1).

I've also verified that the proxy will seamlessly create the specified AWS Secret if it does not yet exist, using the Terraform config in [PR #2 of michaelstepner/email-oauth2-proxy-aws](https://github.com/michaelstepner/email-oauth2-proxy-aws/pull/2).

This lists the various cases that are handled, along with the associated debug log error messages (`--debug` enabled) tested manually on commit 15f049c:

- ✅ Proxy uses AWS-SM (AWS Secrets Manager) when secret is already created and populated
- ✅ Proxy uses AWS-SM when secret does not yet exist, is specified with name, and IAM has permission to create it
- ✅ Warning: `boto3` is not installed
    ```
    2023-01-09 11:33:02: Failed to load AWS SDK (boto3) - "aws_secret" is specified in config file for at least one account; have you installed dependencies from `requirements-aws-secrets.txt`?
    2023-01-09 11:33:02: AWS SDK load error: ModuleNotFoundError("No module named 'boto3'")
    2023-01-09 11:33:02: Unable to get AWS client; cannot fetch OAuth 2.0 tokens from AWS Secrets Manager
    2023-01-09 11:33:02: Starting SMTP server at IP:PORT (TLS) proxying smtp.office365.com:587 (STARTTLS)
    2023-01-09 11:33:02: Initialised Email OAuth 2.0 Proxy - listening for authentication requests. Connect your email client to begin
    ```
- ✅ Warning: IAM user does not have `secretsmanager:GetSecretValue` permissions
    ```
    2023-01-09 11:39:07: Fetching AWS Secret "test1" - access denied: does the IAM user have the "secretsmanager:GetSecretValue" permission?
    2023-01-09 11:39:07: AccessDeniedException: User: arn:aws:iam::000000000000:user/email_oauth2_proxy is not authorized to perform: secretsmanager:GetSecretValue on resource: test1 because no identity-based policy allows the secretsmanager:GetSecretValue action
    2023-01-09 11:39:07: Starting SMTP server at IP:PORT (TLS) proxying smtp.office365.com:587 (STARTTLS)
    2023-01-09 11:39:07: Initialised Email OAuth 2.0 Proxy - listening for authentication requests. Connect your email client to begin
    ```
- ✅ Warning: secret does not exist, cannot create a secret with a specific ARN
    ```
    2023-01-09 13:50:40: Fetching AWS Secret "arn:aws:secretsmanager:us-east-1:000000000000:secret:MySecret-a1b2c3d4e5" - secret does not exist or does not contain a secret value
    2023-01-09 13:50:40: ResourceNotFoundException: Secrets Manager can't find the specified secret.
    [...]
    2023-01-09 13:52:10: Failed to store OAuth 2.0 tokens in AWS Secret "arn:aws:secretsmanager:us-east-1:000000000000:secret:MySecret-a1b2c3d4e5" - the secret does not exist, and it is not possible to create a secret with a specific ARN
    2023-01-09 13:52:10: ResourceNotFoundException: Secrets Manager can't find the specified secret.
    ```
- ✅ Warning: secret does not exist, IAM user does not have `secretsmanager:CreateSecret` permissions
    ```
    2023-01-09 11:47:35: Fetching AWS Secret "test2" - secret does not exist or does not contain a secret value
    2023-01-09 11:47:35: ResourceNotFoundException: Secrets Manager can't find the specified secret.
    [...]
    2023-01-09 11:47:56: AWS Secret "test2" does not exist - attempting to create it
    2023-01-09 11:47:56: ResourceNotFoundException: Secrets Manager can't find the specified secret.
    2023-01-09 11:47:56: Failed to store OAuth 2.0 tokens in AWS Secret "test2" - access denied: does the IAM user have the "secretsmanager:CreateSecret" permission?
    2023-01-09 11:47:56: AccessDeniedException: User: arn:aws:iam::000000000000:user/email_oauth2_proxy is not authorized to perform: secretsmanager:CreateSecret on resource: test2 because no identity-based policy allows the secretsmanager:CreateSecret action
    ```
- ✅ Warning: secret does not exist, IAM user does not have `secretsmanager:PutSecretValue` permissions
    ```
    2023-01-09 13:55:51: Fetching AWS Secret "test2" - secret does not exist or does not contain a secret value
    2023-01-09 13:55:51: ResourceNotFoundException: Secrets Manager can't find the specified secret.
    [...]
    2023-01-09 13:57:38: Failed to store OAuth tokens in AWS Secret "test2" - access denied: does the IAM user have the "secretsmanager:PutSecretValue" permission?
    2023-01-09 13:57:38: AccessDeniedException: User: arn:aws:iam::000000000000:user/email_oauth2_proxy is not authorized to perform: secretsmanager:PutSecretValue on resource: test2 because no identity-based policy allows the secretsmanager:PutSecretValue action
    ```
- ✅ Warning: secret exists, IAM user does not have `secretsmanager:PutSecretValue` permissions
    ```
    2023-01-09 11:45:27: Failed to store OAuth tokens in AWS Secret "test1" - access denied: does the IAM user have the "secretsmanager:PutSecretValue" permission?
    2023-01-09 11:45:27: AccessDeniedException: User: arn:aws:iam::000000000000:user/email_oauth2_proxy is not authorized to perform: secretsmanager:PutSecretValue on resource: test1 because no identity-based policy allows the secretsmanager:PutSecretValue action
    ```


emailproxy.py
```diff
@@ -279,6 +289,58 @@ def _load():
...
+            if 'boto3' not in sys.modules:
+                Log.error('Error: client configuration for some accounts includes an aws_secret'
+                          ' parameter, but the app has been loaded without the --aws-secrets option.')
+                raise ValueError("--aws-secrets must be specified if account configuration includes aws_secret")
+
+            # Create dict of AWS Secret IDs across all accounts
+            aws_secrets = dict.fromkeys([AppConfig._PARSER[account]['aws_secret'] for account in accounts_using_aws_secret], {})
+
+            # Download AWS Secrets
+            aws_client = boto3.client('secretsmanager')
```
simonrob, line 303:
```suggestion
            try:
                aws_client = boto3.client('secretsmanager')
            except botocore.exceptions.BotoCoreError:
                Log.error('Unable to load AWS Secrets Manager - please make sure you have set up the AWS SDK for Python')
                return
```

emailproxy.py
```diff
@@ -324,8 +386,44 @@ def add_account(username):
...
+                aws_secrets = dict.fromkeys([appconfig_to_save[account]['aws_secret'] for account in accounts_using_aws_secret], {})
+
+                # Populate dict with OAuth tokens from each account
+                for account in accounts_using_aws_secret:
+                    aws_secrets[appconfig_to_save[account]['aws_secret']][account] = { key : appconfig_to_save[account][key] for key in TOKEN_KEYS }
+                    for key in TOKEN_KEYS:
+                        appconfig_to_save.remove_option(account, key)
+
+                # Update AWS Secrets
+                aws_client = boto3.client('secretsmanager')
```
simonrob, line 406:
```suggestion
                try:
                    aws_client = boto3.client('secretsmanager')
                except botocore.exceptions.BotoCoreError:
                    Log.error('Unable to load AWS Secrets Manager - please make sure you have set up the AWS SDK for Python')
                    return
```

emailproxy.py
```diff
@@ -91,6 +91,16 @@ class NSObject:
...
 del no_gui_parser
 
+# by default the proxy stores the OAuth 2.0 tokens in a local config file, but it can optionally store them
+# remotely in AWS Secrets Manager. import the dependency only if this option is specified.
+aws_secrets_parser = argparse.ArgumentParser()
+aws_secrets_parser.add_argument('--aws-secrets', action='store_true')
+
+if aws_secrets_parser.parse_known_args()[0].aws_secrets:
+    import boto3
+    from botocore.exceptions import ClientError
```
simonrob, line 101: For consistency I'd `import botocore.exceptions` then use the full package name (not idiomatic Python, but consistent with the rest of the project)

emailproxy.py
```diff
@@ -279,6 +289,58 @@ def _load():
             AppConfig._GLOBALS = configparser.SectionProxy(AppConfig._PARSER, APP_SHORT_NAME)
         AppConfig._SERVERS = [s for s in config_sections if CONFIG_SERVER_MATCHER.match(s)]
         AppConfig._ACCOUNTS = [s for s in config_sections if '@' in s]
+
+        def fetch_aws_secrets():
+            if 'boto3' not in sys.modules:
+                Log.error('Error: client configuration for some accounts includes an aws_secret'
+                          ' parameter, but the app has been loaded without the --aws-secrets option.')
+                raise ValueError("--aws-secrets must be specified if account configuration includes aws_secret")
```
simonrob, line 297: I'm a little reluctant to exit the proxy when errors or misconfigurations occur (both here and other places) unless they are absolutely critical. Perhaps just return here without loading?

simonrob (commented):
Thanks for this contribution! I've had a quick look and added some initial thoughts/suggestions.

michaelstepner:
> Thanks for this contribution! I've had a quick look and added some initial thoughts/suggestions.

@simonrob thanks for the careful review and thoughtful feedback! I plan to iteratively revise & respond to each point over the holidays, and will tag you in a comment when I've finished the first round of revisions.

emailproxy.py
```diff
@@ -279,6 +289,58 @@ def _load():
...
+            # Create dict of AWS Secret IDs across all accounts
+            aws_secrets = dict.fromkeys([AppConfig._PARSER[account]['aws_secret'] for account in accounts_using_aws_secret], {})
+
+            # Download AWS Secrets
+            aws_client = boto3.client('secretsmanager')
+            for secret_id in aws_secrets:
+                try:
+                    get_secret_value_response = aws_client.get_secret_value(SecretId=secret_id)
+                except ClientError as err_getsecret:
+                    if err_getsecret.response['Error']['Code'] == 'ResourceNotFoundException':
```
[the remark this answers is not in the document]
michaelstepner: Resolved in 39e6751

emailproxy.py
```diff
@@ -279,6 +289,58 @@ def _load():
...
+                        if err_getsecret.response['Error']['Message'] == "Secrets Manager can't find the specified secret.":
+                            if secret_id.startswith('arn:'):
+                                Log.error('Error: AWS Secret "%s"'
+                                          ' does not exist, cannot create secret with specific ARN.' % (secret_id))
+                                raise err_getsecret
+                            else:
+                                Log.info('Warning: AWS Secret "%s" does not exist, attempting to create it' % (secret_id))
+                                try:
+                                    create_secret_value_response = aws_client.create_secret(
```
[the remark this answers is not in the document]
michaelstepner:
```
                                Log.info('Warning: AWS Secret "%s" does not exist, attempting to create it' % (secret_id))
                                try:
                                    create_secret_value_response = aws_client.create_secret(
                                        Name=secret_id,
                                        ForceOverwriteReplicaSecret=False)
                                except botocore.exceptions as err_createsecret:
                                    if err_createsecret.response['Error']['Code'] == 'AccessDeniedException':
                                        Log.error('Error: could not create secret, does your IAM user have'
                                                  ' "secretsmanager:CreateSecret" permissions?')
                                    raise err_createsecret
```

The outcome of checking for success/failure here is solely to write a helpful error message in the log.

Would you like me to simplify the code and let the AccessDeniedException thrown by Python serve as the error message here? The "helpful" error message adds the additional context of diagnosing the likely issue and the likely fix.

emailproxy.py
```diff
@@ -324,8 +386,44 @@ def add_account(username):
...
+                for account in accounts_using_aws_secret:
+                    aws_secrets[appconfig_to_save[account]['aws_secret']][account] = { key : appconfig_to_save[account][key] for key in TOKEN_KEYS }
+                    for key in TOKEN_KEYS:
+                        appconfig_to_save.remove_option(account, key)
+
+                # Update AWS Secrets
+                aws_client = boto3.client('secretsmanager')
+                for secret_id in aws_secrets:
+                    try:
+                        response = aws_client.put_secret_value(
```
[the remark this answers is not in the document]
michaelstepner: You're right, this is unnecessarily verbose. Fixed in e0b0925

emailproxy.py
```diff
@@ -2368,6 +2468,8 @@ def get_script_start_command(self):
             script_command.append('--local-server-auth')
         if self.args.config_file:
             script_command.extend(['--config-file', CONFIG_FILE_PATH])
+        if self.args.aws_secrets:
+            script_command.extend(['--aws-secrets', CONFIG_FILE_PATH])
```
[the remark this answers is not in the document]
michaelstepner: Thanks for catching this, fixed in e0b0925

emailproxy.py
michaelstepner: @simonrob I understand your reluctance! Let's make a conscious decision about the expected behaviour when:

- the config for an account specifies `aws_secret:`, so the user has indicated that they want to store their tokens in AWS Secrets Manager.
- the user has called the proxy without the `--aws-secrets` argument, so `boto3` isn't loaded and the proxy cannot use the AWS Secrets Manager.

I see three potential behaviours:
1. Throw an error explaining the issue and exit.
2. Log a warning explaining the issue and then run the proxy. Cache the tokens in memory, but do not write them to the local config file.
3. Log a warning explaining the issue and then run the proxy. Write the tokens to the local config file (as if the user had not specified `aws_secret:`.

In my personal view: 3 is a bad solution, because it defies a user configuration which says "don't store the tokens in my local config file." Options 1 and 2 both seem reasonable to me.

I would lean toward solution 1 (immediately throw an error when the proxy is launched and exit, as currently implemented) because this issue is only going to be encountered by advanced users who are setting up a custom config after reading the docs. They haven't quite gotten their config/command right, because they are instructing the proxy to use AWS Secrets but haven't passed the argument to do so. I think an error in this case is most helpful, since it gives the user immediate feedback to correct their custom config/command.

Solution 2 meets your objective to keep the proxy running whenever possible. I think it is more likely to result in confusion: there will be cases where a user is inattentive, doesn't notice the warning in the logs, the proxy is working fine, and then hours/days/weeks later they wonder why the tokens aren't stored in AWS Secrets Manager in a scenario where they are forced to reauthenticate. They do some debugging and notice the warning about the misconfiguration.

In sum: in this scenario, a user has misconfigured an advanced setting, and I favour solution 1 because it gives immediate feedback prompting them to fix the configuration. But I'd be happy to switch to whatever solution you prefer!

emailproxy.py
```diff
@@ -279,6 +289,58 @@ def _load():
...
+            aws_secrets = dict.fromkeys([AppConfig._PARSER[account]['aws_secret'] for account in accounts_using_aws_secret], {})
+
+            # Download AWS Secrets
+            aws_client = boto3.client('secretsmanager')
+            for secret_id in aws_secrets:
+                try:
+                    get_secret_value_response = aws_client.get_secret_value(SecretId=secret_id)
+                except ClientError as err_getsecret:
+                    if err_getsecret.response['Error']['Code'] == 'ResourceNotFoundException':
+                        if err_getsecret.response['Error']['Message'] == "Secrets Manager can't find the specified secret.":
```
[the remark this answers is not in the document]
michaelstepner: I wish there were, but in my testing I found that the error returned by the AWS API was identical (same HTTP code, same exception) in all values except the message.

Once we resolve the other issues, I'll do some more verbose testing and paste the full contents of the error returned by the AWS API here.

emailproxy.py
```diff
@@ -279,6 +289,58 @@ def _load():
...
+                                          ' does not exist, cannot create secret with specific ARN.' % (secret_id))
+                                raise err_getsecret
+                            else:
+                                Log.info('Warning: AWS Secret "%s" does not exist, attempting to create it' % (secret_id))
+                                try:
+                                    create_secret_value_response = aws_client.create_secret(
+                                        Name=secret_id,
+                                        ForceOverwriteReplicaSecret=False)
+                                except ClientError as err_createsecret:
+                                    if err_createsecret.response['Error']['Code'] == 'AccessDeniedException':
```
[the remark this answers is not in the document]
michaelstepner: Unfortunately I don't believe the error returned by the AWS API is any more specific than this.

Once we resolve the other issues, I'll do some more verbose testing and paste the full contents of the error returned by the AWS API here.

emailproxy.py
michaelstepner: @simonrob I think your proposal here is related to your feedback regarding line 294:

> I'm a little reluctant to exit the proxy when errors or misconfigurations occur (both here and other places) unless they are absolutely critical. Perhaps just return here without loading?

In which case I'll update this code chunk based on the resolution to our conversation in that thread.

If I'm misinterpreting your intention here, let me know!

emailproxy.py
michaelstepner: @simonrob I think your proposal here is related to your feedback regarding line 294:

> I'm a little reluctant to exit the proxy when errors or misconfigurations occur (both here and other places) unless they are absolutely critical. Perhaps just return here without loading?

In which case I'll update this code chunk based on the resolution to our conversation in that thread.

If I'm misinterpreting your intention here, let me know!

emailproxy.py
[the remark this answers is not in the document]
michaelstepner: Good idea. I think your proposed syntax is not quite right (thanks to AWS for using confusing class names), but something equivalent will work. I will test and update in my next submission.

Refs: [How to handle errors with boto3?](https://stackoverflow.com/questions/33068055/how-to-handle-errors-with-boto3)

```
>>> from pprint import pprint
>>> import boto3
>>> iam = boto3.client('iam')
>>> pprint([e for e in dir(iam.exceptions) if e.endswith('Exception')])
['ConcurrentModificationException',
 'CredentialReportExpiredException',
 'CredentialReportNotPresentException',
 'CredentialReportNotReadyException',
 'DeleteConflictException',
 'DuplicateCertificateException',
 'DuplicateSSHPublicKeyException',
 'EntityAlreadyExistsException',
 'EntityTemporarilyUnmodifiableException',
 'InvalidAuthenticationCodeException',
 'InvalidCertificateException',
 'InvalidInputException',
 'InvalidPublicKeyException',
 'InvalidUserTypeException',
 'KeyPairMismatchException',
 'LimitExceededException',
 'MalformedCertificateException',
 'MalformedPolicyDocumentException',
 'NoSuchEntityException',
 'PasswordPolicyViolationException',
 'PolicyEvaluationException',
 'PolicyNotAttachableException',
 'ReportGenerationLimitExceededException',
 'ServiceFailureException',
 'ServiceNotSupportedException',
 'UnmodifiableEntityException',
 'UnrecognizedPublicKeyEncodingException']
```

emailproxy.py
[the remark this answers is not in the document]
michaelstepner: A little more context after I did some research. Error handling with boto3 is a mess, and surprisingly it's Amazon's fault not mine. I'll do some testing to see if I can distinguish the two types of errors using `aws_client.exceptions`, but if not then the error message is the best that can be done...and follows the official AWS docs recommendation 🤦 

Refs:
- [The official docs recommend parsing the error message.](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/error-handling.html#discerning-useful-information-from-error-responses)

  > *Exception class and error message* - You can use this data to build logic around, or in response to, these errors and exceptions.

* [Stack Overflow: How to handle errors with boto3?](https://stackoverflow.com/questions/33068055/how-to-handle-errors-with-boto3)
* [Stack Overflow: Properly catch boto3 Errors](https://stackoverflow.com/questions/46174385/properly-catch-boto3-errors)

emailproxy.py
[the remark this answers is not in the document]
michaelstepner: I'll do some testing to see if I can distinguish the two types of errors using `aws_client.exceptions`, but if not then I think this is the best that can be done. The current code is canonical boto3 error parsing, which is a mess as discussed in other threads.

michaelstepner:
@simonrob I did a pass through all your comments. Thanks again for the feedback!

- I fixed a few minor issues and marked them resolved.
- I asked a few clarifying questions for you.
- I did some more research about error handling with `boto3` and learned that the messiness of my first draft is not just a case of my inexperience. It's quite poorly designed and widely complained about.
  - I'm going to make some revisions after testing, but it's not going to end up as tidy as one would hope.

I will await your feedback on the clarifying questions before tackling the items that require more extensive testing with the AWS SDK.

emailproxy.py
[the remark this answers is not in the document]
simonrob: Thanks for looking into this – I think it will be much better if this code can be less reliant on exact error message text.

emailproxy.py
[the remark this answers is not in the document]
simonrob: No, I think the useful error message is a good thing. But `create_secret_value_response` itself isn't actually used. Is there a return code in that response that needs checking, or does the lack of an error imply success?

emailproxy.py
simonrob: Thanks for the summary of options. I do agree with your reasoning in the first case, but I wonder what the best action would be if the sequence of events is, say:

1: Load the proxy without `--aws-secrets` with a set of accounts that do not have any `aws_secret` values
2: Some time later, edit the configuration file to add `aws_secret` values to a few accounts
3: Reload the configuration file (from the GUI) or send `SIGHUP`

emailproxy.py
```diff
@@ -91,6 +91,16 @@ class NSObject:
             pass
 del no_gui_parser
 
+# by default the proxy stores the OAuth 2.0 tokens in a local config file, but it can optionally store them
+# remotely in AWS Secrets Manager. import the dependency only if this option is specified.
+aws_secrets_parser = argparse.ArgumentParser()
```
simonrob, line 96: Exclude this ArgumentParser from `--help`

```suggestion
aws_secrets_parser = argparse.ArgumentParser(add_help=False)
```

requirements-aws-secrets.txt
```diff
@@ -0,0 +1,14 @@
+# this file contains the requirements to store OAuth 2.0 tokens remotely in AWS Secrets Manager.
+# these are only necessary when specifying the '--aws-secrets' argument.
+boto3
+
+# Note that to use AWS Secrets Manager, you must also set up authentication
+# credentials for your AWS account:
```
simonrob, line 6:
```suggestion

# include core requirements (separated in order to support usage without GUI-only dependencies)
-r requirements-no-gui.txt

# Note that to use AWS Secrets Manager you must also set up authentication credentials for your AWS account:
```

requirements-aws-secrets.txt
```diff
@@ -0,0 +1,14 @@
...
+# https://boto3.amazonaws.com/v1/documentation/api/latest/guide/quickstart.html#configuration
+#
+# The minimal required permissions for each AWS Secret used are:
+#   "secretsmanager:GetSecretValue"
+#   "secretsmanager:PutSecretValue"
+#
+# If an AWS Secret does not yet exist, the following permissions are also required:
+#   "secretsmanager:CreateSecret"
```
simonrob, line 14:
```suggestion
#   "secretsmanager:CreateSecret"
boto3

```

simonrob (commented):
Here I've suggested a minor reorganisation of the requirements file to include the core no-GUI requirements so that this file can be relied upon to install all dependencies.

emailproxy.py
michaelstepner: @simonrob That's an excellent question – I wasn't aware that the config could be reloaded on the fly until today. I liked the idea of exiting the proxy with an error immediately upon launch, but I like it much less if it happens in scenarios that aren't immediately upon launch.

Given your point, I think the best course of action is Solution 2: log a warning explaining the issue, continue running the proxy, cache the tokens in memory, but do not write them to the local config file. The proxy keeps running, but will require re-authentication when it is next restarted.

I'll go ahead and implement that behaviour in my next revision, and test it along with testing the AWS errors to see if we can get a more precise solution there.

michaelstepner:
@simonrob I've substantially refactored the code in commits 5b7d63b and 2c9c6be in order to:

- Address all of your feedback in the code review thus far
- Eliminate the proposed `--aws-secrets` argument, which was no longer necessary, so removing it simplifies configuration for the user.
- Isolate the new code which implements the secrets manager within clearly-named methods in the AppConfig class
    - This feature only interacts with existing code via simple checks in `AppConfig._load()` and `AppConfig.save()` to see if AWS Secrets Manager functions should be called.
- Provide more informative error messages

I'm much happier with this new design: I believe the code is more robust and more maintainable.

It's ready for a new round of code review when you have a chance.

emailproxy.py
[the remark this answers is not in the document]
michaelstepner: In my latest revisions of the code for this PR, I've continued to catch `botocore.exceptions.ClientError` and check the error code based on the [advice from the official boto3 docs](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/error-handling.html#parsing-error-responses-and-catching-exceptions-from-aws-services):

> Catching exceptions through `ClientError` and parsing for error codes is still the best way to catch **all** service-side exceptions and errors.

However my new code does not parse the error message, and provides guidance when the ClientError is not one that I specifically catch and provide instructive feedback: "... unexpected error, see debug logs for details"

emailproxy.py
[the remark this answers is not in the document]
michaelstepner: Resolved by moving the creation of the secret to the `save` step rather than the `load` step, at which point we don't need to distinguish between "no secret" vs "secret exists but no secret value"

requirements-aws-secrets.txt
```diff
@@ -0,0 +1,22 @@
...
+#         "secretsmanager:GetSecretValue"
+#         "secretsmanager:PutSecretValue"
+#
+#     - If an AWS Secret does not yet exist, the following permissions are also required:
+#         "secretsmanager:CreateSecret"
+
+boto3
+
+# include core requirements (separated in order to support usage without GUI-only dependencies)
+-r requirements-no-gui.txt
```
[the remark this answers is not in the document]
simonrob: Minor edits to match existing linting settings and styles.

emailproxy.py
```diff
@@ -337,9 +341,141 @@ def add_account(username):
...
+            Log.error('Failed to load AWS SDK - "aws_secret" is specified in config file for some accounts, have you installed requirements-aws-secrets.txt?')
+            Log.debug(Log.error_string(e))
+            return
+        else:
+            return aws_client
+
+    @staticmethod
+    def aws_secrets_debuglog_boto3error(err):
+        Log.debug(err.response.get("Error", {}).get("Code") + ":",
+                  err.response.get("Error", {}).get("Message"))
```
simonrob, line 382: Why not just `err.response['Error']['Code']` and `err.response['Error']['Message']`?

What is the default value of `{}` for here?

emailproxy.py
```diff
@@ -337,9 +341,141 @@ def add_account(username):
...
+                else:
+                    aws_secrets[secret_id] = json.loads(get_secret_value_response['SecretString'])
+
+            # Update local config in memory
+            for account in AppConfig.aws_secrets_accounts():
+                if account in aws_secrets[AppConfig._PARSER[account]['aws_secret']]:
+                    for key in ['token_salt','access_token','access_token_expiry','refresh_token']:
+                        AppConfig._PARSER.set(account, key, aws_secrets[AppConfig._PARSER[account]['aws_secret']][account][key])
+        else:
+            Log.error('Failed to fetch OAuth 2.0 tokens from AWS Secrets Manager')
```
simonrob, line 412:
```suggestion
    def _aws_secrets_fetch():
        aws_client = AppConfig._aws_secrets_boto3_client()
        if aws_client:
            # get the values of all AWS Secrets referenced in the configuration file
            aws_secrets = AppConfig._aws_secrets_ids()
            for secret_id in aws_secrets:
                try:
                    get_secret_value_response = aws_client.get_secret_value(SecretId=secret_id)

                except botocore.exceptions.ClientError as err_get_secret:
                    if err_get_secret.response['Error']['Code'] == 'ResourceNotFoundException':
                        Log.info('Fetching AWS Secret "%s"' % secret_id, '- secret does not exist or does not contain',
                                 'a secret value')
                    elif err_get_secret.response['Error']['Code'] == 'AccessDeniedException':
                        Log.error('Fetching AWS Secret "%s"' % secret_id, '- access denied: does the IAM user have the',
                                  '"secretsmanager:GetSecretValue" permission?')
                    else:
                        Log.error('Fetching AWS Secret "%s"' % secret_id, '- unexpected error; see the proxy\'s debug',
                                  'mode log for details')
                    AppConfig._aws_secrets_debug_log_boto3_error(err_get_secret)

                else:
                    aws_secrets[secret_id] = json.loads(get_secret_value_response['SecretString'])

            # update local proxy configuration in memory
            for account in AppConfig._aws_secrets_accounts():
                account_secret = AppConfig._PARSER.get(account, 'aws_secret')
                if account in aws_secrets[account_secret]:
                    for key in AppConfig._SECRETS_MANAGER_KEYS:
                        AppConfig._PARSER.set(account, key, aws_secrets[account_secret][account][key])
        else:
            Log.error('Unable to get AWS client; cannot fetch OAuth 2.0 tokens from AWS Secrets Manager')
```

Minor edits for clarity/consistency.

emailproxy.py
```diff
@@ -337,9 +341,141 @@ def add_account(username):
...
+                        Log.error('Failed to store OAuth tokens in AWS Secret "%s" - access denied, does IAM user have "secretsmanager:PutSecretValue" permissions?' % (secret_id))
+                        AppConfig.aws_secrets_debuglog_boto3error(err_putsecret)
+                    else:
+                        Log.error('Failed to store OAuth tokens in AWS Secret "%s" - unexpected error, see debug logs for details' % (secret_id))
+                        AppConfig.aws_secrets_debuglog_boto3error(err_putsecret)
+        else:
+            Log.error('Failed to store OAuth 2.0 tokens in AWS Secrets Manager')
 
+        # Return copy of config without OAuth 2.0 tokens to write to disk
+        return appconfig_to_save
```
simonrob, line 481:
```suggestion
    def _aws_secrets_save():
        # a deep copy of the configuration file to be saved locally, *without* OAuth 2.0 tokens (_SECRETS_MANAGER_KEYS)
        appconfig_to_save = configparser.ConfigParser()
        appconfig_to_save.read_dict(AppConfig._PARSER)

        # remove OAuth 2.0 tokens from each account, transferring them into a local dict of referenced AWS Secrets
        aws_secrets = AppConfig._aws_secrets_ids()
        for account in AppConfig._aws_secrets_accounts():
            account_secret = appconfig_to_save.get(account, 'aws_secret')
            aws_secrets[account_secret][account] = {key: appconfig_to_save.get(account, key) for key in
                                                    AppConfig._SECRETS_MANAGER_KEYS}
            for key in AppConfig._SECRETS_MANAGER_KEYS:
                appconfig_to_save.remove_option(account, key)

        # save the extracted AWS Secrets, attempting to create new secrets if they do not already exist
        aws_client = AppConfig._aws_secrets_boto3_client()
        if aws_client:
            for secret_id in aws_secrets:
                try:
                    aws_client.put_secret_value(SecretId=secret_id, SecretString=json.dumps(aws_secrets[secret_id]))

                except botocore.exceptions.ClientError as err_put_secret:
                    if err_put_secret.response['Error']['Code'] == 'ResourceNotFoundException':
                        if secret_id.startswith('arn:'):
                            Log.error('Failed to store OAuth 2.0 tokens in AWS Secret "%s"' % secret_id, '- the Secret',
                                      'does not exist, and it is not possible to create a secret with a specific ARN')
                            AppConfig._aws_secrets_debug_log_boto3_error(err_put_secret)

                        else:
                            Log.error('AWS Secret "%s"' % secret_id, 'does not exist - attempting to create it')
                            AppConfig._aws_secrets_debug_log_boto3_error(err_put_secret)

                            try:
                                aws_client.create_secret(Name=secret_id, ForceOverwriteReplicaSecret=False)

                            except botocore.exceptions.ClientError as err_create_secret:
                                if err_create_secret.response['Error']['Code'] == 'AccessDeniedException':
                                    Log.error('Failed to store OAuth 2.0 tokens in AWS Secret "%s"' % secret_id,
                                              '- access denied: does the IAM user have the',
                                              '"secretsmanager:CreateSecret" permission?')
                                else:
                                    Log.error('Failed to store OAuth 2.0 tokens in AWS Secret "%s"' % secret_id,
                                              '- attempt to create failed with unexpected error; see the proxy\'s',
                                              'debug mode log for details')
                                AppConfig._aws_secrets_debug_log_boto3_error(err_create_secret)

                            else:
                                Log.info('Created AWS Secret "%s"' % secret_id)
                                try:
                                    aws_client.put_secret_value(SecretId=secret_id,
                                                                SecretString=json.dumps(aws_secrets[secret_id]))

                                except botocore.exceptions.ClientError as err_put_secret_after_create:
                                    if err_put_secret_after_create.response['Error']['Code'] == 'AccessDeniedException':
                                        Log.error('Failed to store OAuth 2.0 tokens in AWS Secret "%s"' % secret_id,
                                                  '- access denied: does the IAM user have the',
                                                  '"secretsmanager:PutSecretValue" permission?')
                                    else:
                                        # TODO: missing format string
                                        Log.error('Failed to store OAuth 2.0 tokens in AWS Secret "%s"' % secret_id,
                                                  '- unexpected error; see the proxy\'s debug mode log for details')
                                    AppConfig._aws_secrets_debug_log_boto3_error(err_put_secret_after_create)

                    elif err_put_secret.response['Error']['Code'] == 'AccessDeniedException':
                        Log.error('Failed to store OAuth tokens in AWS Secret "%s"' % secret_id, '- access denied:',
                                  'does the IAM user have the "secretsmanager:PutSecretValue" permission?')
                        AppConfig._aws_secrets_debug_log_boto3_error(err_put_secret)

                    else:
                        Log.error('Failed to store OAuth tokens in AWS Secret "%s"' % secret_id, '- unexpected error;',
                                  ' see the proxy\'s debug mode log for details')
                        AppConfig._aws_secrets_debug_log_boto3_error(err_put_secret)
        else:
            Log.error('Unable to get AWS client; cannot store OAuth 2.0 tokens in AWS Secrets Manager')

        # the updated configuration without OAuth 2.0 tokens (see AppConfig._SECRETS_MANAGER_KEYS)
        return appconfig_to_save
```

Minor edits for consistency/clarity (and fixed a few missing format strings).

emailproxy.py
```diff
@@ -337,9 +341,141 @@ def add_account(username):
     @staticmethod
     def save():
         if AppConfig._LOADED:
+            if AppConfig.aws_secrets_accounts():
+                config_to_write = AppConfig.aws_secrets_save()
```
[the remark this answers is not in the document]
michaelstepner: Agreed!

emailproxy.py
michaelstepner: Agreed, the convoluted code is a symptom of copy pasting. Will revise to be clearer.

emailproxy.py
michaelstepner: Will apply further minor changes:
- `the Secret` -> `the secret` on line 439
- Remove `# TODO: missing format string` on line 473

michaelstepner (commented):
@simonrob thank you for putting in another round of detailed review and feedback! I am accepting all of your suggestions and will make a couple of my own.

emailproxy.py
```diff
@@ -337,9 +344,158 @@ def add_account(username):
...
+            Log.error('Failed to load AWS SDK (boto3) - "aws_secret" is specified in config file for at least one',
+                      'account; have you installed dependencies from `requirements-aws-secrets.txt`?')
+            Log.debug('AWS SDK load error:', Log.error_string(e))
+        else:
+            return aws_client
+
+    @staticmethod
+    def _aws_secrets_debug_log_boto3_error(err):
+        Log.debug(err.response.get("Error", {}).get("Code") + ":",
+                  err.response.get("Error", {}).get("Message"))
```
michaelstepner, line 385:
```suggestion
        Log.debug(err.response['Error']['Code'] + ":",
                  err.response['Error']['Code'])
```

emailproxy.py
```diff
@@ -337,9 +344,158 @@ def add_account(username):
...
+
+                                except botocore.exceptions.ClientError as err_put_secret_after_create:
+                                    if err_put_secret_after_create.response['Error']['Code'] == 'AccessDeniedException':
+                                        Log.error('Failed to store OAuth 2.0 tokens in AWS Secret "%s"' % secret_id,
+                                                  '- access denied: does the IAM user have the',
+                                                  '"secretsmanager:PutSecretValue" permission?')
+                                    else:
+                                        # TODO: missing format string
+                                        Log.error('Failed to store OAuth 2.0 tokens in AWS Secret "%s"' % secret_id,
+                                                  '- unexpected error; see the proxy\'s debug mode log for details')
```
michaelstepner, line 482:
```suggestion
                                        Log.error('Failed to store OAuth 2.0 tokens in AWS Secret "%s"' % secret_id,
                                                  '- unexpected error; see the proxy\'s debug mode log for details')
```

Removing stale TODO comment

emailproxy.py
```diff
@@ -337,9 +344,158 @@ def add_account(username):
...
+        if aws_client:
+            for secret_id in aws_secrets:
+                try:
+                    aws_client.put_secret_value(SecretId=secret_id, SecretString=json.dumps(aws_secrets[secret_id]))
+
+                except botocore.exceptions.ClientError as err_put_secret:
+                    if err_put_secret.response['Error']['Code'] == 'ResourceNotFoundException':
+                        if secret_id.startswith('arn:'):
+                            Log.error('Failed to store OAuth 2.0 tokens in AWS Secret "%s"' % secret_id, '- the Secret',
+                                      'does not exist, and it is not possible to create a secret with a specific ARN')
```
michaelstepner, line 447:
```suggestion
                            Log.error('Failed to store OAuth 2.0 tokens in AWS Secret "%s"' % secret_id, '- the secret',
                                      'does not exist, and it is not possible to create a secret with a specific ARN')
```

emailproxy.py
```diff
@@ -337,9 +344,158 @@ def add_account(username):
...
+                config_to_write = AppConfig._PARSER
+
             with open(CONFIG_FILE_PATH, mode='w', encoding='utf-8') as config_output:
-                AppConfig._PARSER.write(config_output)
+                config_to_write.write(config_output)
+
+    @staticmethod
+    def _aws_secrets_accounts():
+    # all accounts that are using AWS Secrets
+        return [a for a in AppConfig._ACCOUNTS if AppConfig._PARSER.get(a, 'aws_secret', fallback=None)]
```
michaelstepner, line 358:
```suggestion
    def _aws_secrets_accounts():
        # all accounts that are using AWS Secrets
        return [a for a in AppConfig._ACCOUNTS if AppConfig._PARSER.get(a, 'aws_secret', fallback=None)]
```

simonrob:
I've approved all suggestions from your latest pass. Let me know if you find any issues running with these edits; otherwise I think this can be merged.

michaelstepner:
> I've approved all suggestions from your latest pass. Let me know if you find any issues running with these edits; otherwise I think this can be merged.

Thanks @simonrob! I'll try to re-run my tests over the weekend so we can merge next week. I really appreciate the time you took to consider and improve my PR.

emailproxy.py
```diff
@@ -337,9 +344,157 @@ def add_account(username):
...
+        # a deep copy of the configuration file to be saved locally, *without* OAuth 2.0 tokens (_SECRETS_MANAGER_KEYS)
+        appconfig_to_save = configparser.ConfigParser()
+        appconfig_to_save.read_dict(AppConfig._PARSER)
+
+        # remove OAuth 2.0 tokens from each account, transferring them into a local dict of referenced AWS Secrets
+        aws_secrets = AppConfig._aws_secrets_ids()
+        for account in AppConfig._aws_secrets_accounts():
+            account_secret = appconfig_to_save.get(account, 'aws_secret')
+            aws_secrets[account_secret][account] = {key: appconfig_to_save.get(account, key) for key in
+                                                    AppConfig._SECRETS_MANAGER_KEYS}
```
simonrob, line 432: One thing we haven't covered so far is what would/should happen if the required keys are not present (either in the secret, or in the proxy's cache).

I don't know if/how this could happen from the AWS side, but the proxy can (and does) delete values when they become invalid, which would cause this method to fail.

One simple solution here would be to use a fallback value in `get()` and skip keys if they have no value (and entire accounts if there are no valid keys). What do you think?

simonrob: Another related question: my assumption is that token renewal is handled via a local copy of the proxy - i.e., you receive an authentication failure when using your remote server, which prompts you to open a local copy and authenticate, updating the AWS secret. Is that correct? If so, how do you trigger a token from AWS Secrets Manager on the remote server update?

emailproxy.py
```diff
@@ -337,9 +344,157 @@ def add_account(username):
...
+
+                else:
+                    aws_secrets[secret_id] = json.loads(get_secret_value_response['SecretString'])
+
+            # update local proxy configuration in memory
+            for account in AppConfig._aws_secrets_accounts():
+                account_secret = AppConfig._PARSER.get(account, 'aws_secret')
+                if account in aws_secrets[account_secret]:
+                    for key in AppConfig._SECRETS_MANAGER_KEYS:
+                        AppConfig._PARSER.set(account, key, aws_secrets[account_secret][account][key])
```
michaelstepner, line 417:
```suggestion
                    for key in AppConfig._SECRETS_MANAGER_KEYS:
                        AppConfig._PARSER.set(account, key, aws_secrets[account_secret][account].get(key, ''))
```

emailproxy.py
michaelstepner:
```suggestion
            aws_secrets[account_secret][account] = {key: appconfig_to_save.get(account, key, '') for key in
                                                    AppConfig._SECRETS_MANAGER_KEYS}
```

emailproxy.py
```diff
@@ -337,9 +344,157 @@ def add_account(username):
...
+        appconfig_to_save.read_dict(AppConfig._PARSER)
+
+        # remove OAuth 2.0 tokens from each account, transferring them into a local dict of referenced AWS Secrets
+        aws_secrets = AppConfig._aws_secrets_ids()
+        for account in AppConfig._aws_secrets_accounts():
+            account_secret = appconfig_to_save.get(account, 'aws_secret')
+            aws_secrets[account_secret][account] = {key: appconfig_to_save.get(account, key) for key in
+                                                    AppConfig._SECRETS_MANAGER_KEYS}
+            for key in AppConfig._SECRETS_MANAGER_KEYS:
+                appconfig_to_save.remove_option(account, key)
```
michaelstepner, line 434: Will `appconfig_to_save.remove_option(account, key)` throw an exception and crash the proxy if the key does not exist?

michaelstepner (commented):
> One thing we haven't covered so far is what would/should happen if the required keys are not present (either in the secret, or in the proxy's cache).
> 
> I don't know if/how this could happen from the AWS side, but the proxy can (and does) delete values when they become invalid, which would cause this method to fail.
> 
> One simple solution here would be to use a fallback value in get() and skip keys if they have no value (and entire accounts if there are no valid keys). What do you think?

I manually tested what would happen if I edited the secret value in AWS Secrets Manager to remove a key (e.g. remove `token_salt`). As you predicted, this causes a crash because we don't have a fallback value.

I agree that a missing key should not cause a crash. The proxy should have the same behaviour as it would if the local config had a missing key.

I've proposed two changes that would set a default value of an empty string `''` when fetching/storing the token keys.

I also posed a question below about whether we need to add additional error handling
logic around `remove_option`.

emailproxy.py
michaelstepner: I addressed the first question about required keys in a review below.

> Another related question: my assumption is that token renewal is handled via a local copy of the proxy - i.e., you receive an authentication failure when using your remote server, which prompts you to open a local copy and authenticate, updating the AWS secret. Is that correct? If so, how do you trigger a token update from AWS Secrets Manager on the remote server?

I think there are 3 options:

1. handle token renewal locally and use SIGHUP to reload config on the remote server.
2. handle token renewal locally and restart the remote server's proxy to reload config
3. handle the token renewal by ssh-ing into the remote server.

emailproxy.py
michaelstepner, line 417:
```suggestion
                    for key in [k for k in AppConfig._SECRETS_MANAGER_KEYS if k in aws_secrets[account_secret][account]]:
                        AppConfig._PARSER.set(account, key, aws_secrets[account_secret][account].get(key))
```

Here's a potentially preferable solution, which avoids the use of an empty string to denote an undefined key.

emailproxy.py
michaelstepner, line 434:
```suggestion
            aws_secrets[account_secret][account] = {}
            for key in [k for k in AppConfig._SECRETS_MANAGER_KEYS if k in appconfig_to_save[account]]:
                aws_secrets[account_secret][account][key] = appconfig_to_save.get(account, key)
                appconfig_to_save.remove_option(account, key)
```
Here's a potentially preferable solution, which avoids the use of an empty string to denote an undefined key.

emailproxy.py
michaelstepner: See potentially preferable solution below that avoids the use of an empty string.

emailproxy.py
michaelstepner: See potentially preferable solution below that avoids the use of an empty string and should avoid throwing an exception due to attempting to remove a key that does not exist.

simonrob:
```suggestion
            aws_secrets[account_secret][account] = {}
            for key in AppConfig._SECRETS_MANAGER_KEYS:
                if appconfig_to_save.has_option(account, key):
                    aws_secrets[account_secret][account][key] = appconfig_to_save.get(account, key)
                    appconfig_to_save.remove_option(account, key)
```

I think I'd prefer using the configparser methods rather than direct dict access – here's an alternative approach

emailproxy.py
simonrob:
```suggestion
                    for key in AppConfig._SECRETS_MANAGER_KEYS:
                        if k in aws_secrets[account_secret][account]:
                            AppConfig._PARSER.set(account, key, aws_secrets[account_secret][account][key])
```

I think for future legibility this is a simpler approach.

emailproxy.py
simonrob: I was wondering whether it is worth triggering `AppConfig._aws_secrets_fetch()` (probably via a more generic method in the AppConfig class) after a login failure for an account using AWS Secrets. This way, a login failure on the server would trigger checking for updated credentials that have been provided via a local login, and there would be no need to log in to the server at all.

emailproxy.py
michaelstepner: Resolving in favour of better solution proposed later.

emailproxy.py
michaelstepner: Resolving in favour of better solution proposed later.

michaelstepner: Agreed your proposal is cleaner, added to commit e11338d

emailproxy.py
michaelstepner: Agreed your proposal is cleaner, added to commit e11338d

emailproxy.py
michaelstepner: > I was wondering whether it is worth triggering AppConfig._aws_secrets_fetch() (probably via a more generic method in the AppConfig class) after a login failure for an account using AWS Secrets. This way, a login failure on the server would trigger checking for updated credentials that have been provided via a local login, and there would be no need to log in to the server at all.

@simonrob that's an excellent idea. Would it make sense to call `AppConfig._load()` to generically reload the config after a login failure? Thus reloading both the local and remote config.

simonrob: There's no actual need to reload the whole file, though that is a potential option (via `AppConfig.reload()`).

Either way, the main question is _when_ to actually do this. The trigger to set up a local copy and provide updated tokens is a login failure, so although [this](https://github.com/michaelstepner/email-oauth2-proxy/blob/e11338d444218fbd7f59b2a4f86d933066c1db24/emailproxy.py#L665) is a natural place, there's no way you could actually set up and authenticate via a local copy in the meantime between this failure and the subsequent no-cache retry.

A more viable option is to reload from AWS [here](https://github.com/michaelstepner/email-oauth2-proxy/blob/e11338d444218fbd7f59b2a4f86d933066c1db24/emailproxy.py#L562), if there is no access token, and then recurse (and exit early) if the token value changes after that reload.

Are there any cost implications to reloading from AWS Secrets? I can imagine an email client failing repeatedly might trigger a lot of requests...

michaelstepner: @simonrob the cost of AWS Secrets Manager is $0.40 per month + $0.05 per 10,000 API calls. So as long as the reload isn't happening in a rapid loop, I don't see a cost concern. If it happened once per minute for an entire 31-day month, it would cost US$0.22.

I'm not familiar with that section of code, so I'd defer to your judgement on the most sensible place to reload.

simonrob: Here's a quick go at a solution (completely untested, and unlikely to be the final approach) that is worth you trying out. Put this after [this line](https://github.com/michaelstepner/email-oauth2-proxy/blob/e11338d444218fbd7f59b2a4f86d933066c1db24/emailproxy.py#L562):

```python
        # try reloading from Secrets Manager to retrieve remotely-updated tokens if they exist 
        if not access_token and username in AppConfig._aws_secrets_accounts():
            AppConfig.reload()
            return OAuth2Helper.get_oauth2_credentials(username, password, recurse_retries=False)
```

Looking at this made me reconsider the `allow_catch_all_accounts` advanced configuration option. For consistency, having an `aws_secret` value in any parent account (e.g., the global [@], or, say [@gmail.com]) should use that secret to store values for all child accounts within that group.  I'll think about how is best to address this.

michaelstepner: Thanks @simonrob! Happy to test that out, although looking at my work week I'm not sure how soon I'll get to it this week.

Since we've completed the core functionality, do you think it would be worthwhile to merge this PR and split the reload functionality into a new PR? Or would you prefer to keep it in a single PR?

michaelstepner: > Looking at this made me reconsider the `allow_catch_all_accounts` advanced configuration option. For consistency, having an `aws_secret` value in any parent account (e.g., the global [@], or, say [@gmail.com]) should use that secret to store values for all child accounts within that group. I'll think about how is best to address this.

The simplest brute solution would be to only support `aws_secret` functionality for fully-specified accounts, and return a warning & ignore the `aws_secret` if it's specified for a catch_all account. Up to you how valuable it would be to get these two features to work together, but I don't foresee using that functionality personally.

michaelstepner: On second thought, I think the functionality already implemented so far will work fine for catch-all accounts (unless I'm mistaken). It's the reload step that is trickier.

simonrob: I think the catch-all functionality is quite valuable, but it should be relatively straightforward to incorporate that into the existing PR. I'll take a look when I get chance.

How did you get on with the reloading suggestion?

michaelstepner: I've had a busy week and haven't had a chance to test the reloading approach you suggested, but I look forward to testing it soon—perhaps this weekend.

Re: catch-all with AWS Secrets, I agree it will be best to integrate! I haven't used catch-all accounts before because I only use the proxy for a single account, but I suppose I can define a catch-all that applies to that single account in order to test.

michaelstepner:
@simonrob based on [@thiagomacieira's question in #33](https://github.com/simonrob/email-oauth2-proxy/issues/33#issuecomment-1397551005), I wonder if you'd like to adjust the way that we're specifying this feature in the config, in order to provide future flexibility for alternative token storage strategies.

Right now, we're enabling AWS Secrets Manager token storage by adding the following to config:

```
aws_secret = *** ARN or name of secret in AWS Secrets Manager ***
```

We could potentially modify the config syntax to "future-proof" it, ex:

```
token_storage = { "type": "aws_secrets_manager", "id": "*** ARN or name of secret in AWS Secrets Manager ***" }
```

In the future, if for example someone implemented token storage in an alternative local file, it could use the same config key, ex:

```
token_storage = { "type": "local", "file": "path/to/local/file" }
```

simonrob:
That's an interesting idea, but I wonder whether it is a little overly complex. Looking at the current implementation, is there actually any need to have a separate AWS Secrets Manager ID for each account entry, or could this be a global setting?

If not, both the AWS feature and the ability to store tokens in a local cache file could as you point out be implemented in a very similar way. I'd probably make this a little more generic, and have the `AppConfig.save()` method strip out everything except the key tokens from a `ConfigParser` deep copy, and then just pass this object to whichever saving method is configured. As it happens, this would also remove any issues with catch-all accounts, which makes things simpler in the AWS case too.

What do you think – could this all be handled via a global configuration parameter?

michaelstepner:
Although I don't personally store the credentials for different accounts in different locations, that's because I have only one account and am using the proxy in a single-user setting. I implemented it as an account-specific parameter because it seemed plausible to me that one might like to have separate access controls for the different accounts.

Suppose I'm running a remote server for two different users, each of whom I'd like to empower to update their own credentials from their local computers. If it's a global config parameter, then each user must have read/write access to each other's tokens. If it's an account-specific config parameter, then each user could have individual access controls to their own secrets.

Do you think the JSON value for `token_storage` is especially complex? It seems like it would be a very small change to the current code, although it allows complete flexibility for the configuration of other token storage options in the future. (Which I suppose could become complex, although they could remain simple.)

simonrob:
I think it becomes overly complex once additional external token stores come into the picture. I'm also not sure how often the sort of situation you've outlined would actually happen - remember each user would also have to have their own copy of the proxy to re-authenticate periodically and then synchronise with the server. Is that likely?

I think it's probably more common that someone running that setup would either be an individual (like yourself) or an organisation that uses one of the O365 methods that removes the need for local OAuth.

If this is made a global setting then everything becomes much simpler (and easier to maintain) - if you want to add a new token store you are simply provided with a ConfigParser object, and you write all of its contents to whatever storage you like. There are no issues with any other proxy features either (except possibly client secret encryption, but it's unlikely you'd be using that in this case).

michaelstepner:
> I'm also not sure how often the sort of situation you've outlined would actually happen - remember each user would also have to have their own copy of the proxy to re-authenticate periodically and then synchronise with the server. Is that likely?

Not sure! In this case we're entirely in the realm of speculation regarding future use cases that neither of us will personally face.

> If this is made a global setting then everything becomes much simpler (and easier to maintain) - if you want to add a new token store you are simply provided with a ConfigParser object, and you write all of its contents to whatever storage you like. There are no issues with any other proxy features either (except possibly client secret encryption, but it's unlikely you'd be using that in this case).

Ultimately you're the maintainer of the package and I will gladly defer to you on the tradeoffs regarding flexibility vs complexity/maintainability.

Do you want to refactor the current PR to make the token storage a global setting?

simonrob:
I spent some time just now thinking about how to balance the complexity, multi-account inheritance, desire for other external token caches and ongoing maintainability. 2e6b641 is what I ended up with – this adds a new `--cache-store` script parameter, removes the ability to specify a different AWS Secret per account, and also adds a new requirement to use a prefix if specifying a name rather than an ARN. But it means that other cache stores can be added far more easily, and in my opinion is a bit more maintainable going forward.

What do you think?

simonrob:
@michaelstepner I'm pretty happy with this new version, but it'd be useful to get your feedback if you have time.

simonrob:
@michaelstepner any thoughts about this?

simonrob:
I plan to merge this next week, so if you have any feedback or issues please do let me know before then.

simonrob:
Thanks @michaelstepner for all the work on this – I think it's time to merge.

simonrob:
It wouldn't have happened at all without your idea and contribution, so no worries at all. Really glad to hear it works well with your tool; thanks again for the work on this.

PR #120, opened 2023-01-10
Allow binding to IPv6 addresses

brianjmurrell:
Allow binding to IPv6.

The system I am on has IPv6 support and this works.  I'm not sure what the implications of such a simple change are with an IPv4 only system.  Do those even exist any more?  :smile:

OK, IPv6 is nowhere near ubiquitous, but aren't most modern systems dual-stack even when they don't have external/local network IPv6 support?

### DIFF
--- a/emailproxy.py
+++ b/emailproxy.py
@@ -1696,7 +1696,7 @@ class OAuth2Proxy(asyncore.dispatcher):
 
     def start(self):
         Log.info('Starting', self.info_string())
-        self.create_socket(socket.AF_INET, socket.SOCK_STREAM)
+        self.create_socket(socket.AF_INET6, socket.SOCK_STREAM)
         self.set_reuse_addr()
         self.bind(self.local_address)
         self.listen(5)


brianjmurrell:
Hrm.  This might only allow binding to IPv6 and refuse connections from IPv4.  Might have to go back to the drawing board on this one.

simonrob:
Thanks for the contribution – yes, the use of asyncore (which is deprecated) makes things like this harder than they should be. The proxy also currently supports Python versions back to 3.6, and `create_server` was added in 3.8. However, this may not matter – [`create_server`](https://github.com/python/cpython/blob/12be23cf3c1301be2c6b8fd4cb2cd35a567d2ea2/Lib/socket.py#L877) doesn't do anything particularly special, and is just a convenience function. I'm happy to review a change that implements this in a way that supports IPv4 and IPv6. (I presume you'd also want to look at [outgoing connections](https://github.com/simonrob/email-oauth2-proxy/blob/a26edc6e1370f399894059c1fe96beee666eda7a/emailproxy.py#L1304)).

simonrob:
I'm going to close this PR because #140 has a more complete solution. It would be great to get your feedback on that approach though.

brianjmurrell:
> I'm going to close this PR because #140 has a more complete solution. It would be great to get your feedback on that approach though.

It seems reasonable but I no longer have the IPv6/dual-stack environment that I was working in.

PR #140, opened 2023-03-01
IPv6 binding with Dual Stack Support

sahilph:
This PR will allow binding to IPv6 addresses. It also has ability to accept both IPv4 and IPv6 connections by creating a Dual Stack bind on supported systems.

### DIFF
--- a/emailproxy.config
+++ b/emailproxy.config
@@ -26,7 +26,9 @@ documentation = Local servers are specified as demonstrated below where, for exa
 	behalf (i.e., do not enable STARTTLS in your client). IMAP STARTTLS and POP STARTTLS are not currently supported.
 	
 	- If the property `local_address` is not specified, its value is assumed to be `localhost`. If required, this
-	parameter can be used to set an IP address or hostname for the proxy to listen on.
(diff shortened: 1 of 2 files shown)


sahilph:
Hello @simonrob , I was just wondering if you had a chance to review this ?

simonrob:
Thanks for the contribution!

I had been investigating this myself offline, and had ended up with a broadly similar method to yours. I've now merged them and pushed to your branch, fixing a few issues with both approaches.

For me (macOS) this works well, and supports IPv6, IPv4 and dual-stack options. Could you test this on your own setup and let me know how you get on?

sahilph:
Hello @simonrob .

Dual Stack initially didn't work as I has set the environment to Python 3.6, because I read #120 and the README and assumed that this was supposed to support Python 3.6+

But when I changed to Python 3.8,  dual stack worked fine.

Single Stack IPv6 (or IPV4) Binding is working fine for Python 3.6 (as well as 3.8)

I see that you have added a comment regarding this in the code, but maybe you could add a note in the README or config file as well that dual-stack will only work with Python 3.8+ as usually users will not go through the code.

simonrob:
Thanks for the review – I've clarified in the configuration file readme. All good to merge now I think.

Issue #152, opened 2023-03-30
Only IPv6 is bound


wtcline-intc:
Since version `2023-03-09`, only the IPv6 interface is being bound.  I tracked this down to commit `966e6e23376afc08c4429e8d6b3b56e23b175565`.  The issue affects a fresh installation of Fedora 37 and Ubuntu 22.04 (minimal install) with the default proxy configuration.

Expected: The proxy continues to bind to IPv4 as before.

Ubuntu 22.04:
```
$ pip3 install -r requirements.txt
$ sudo apt-get install gir1.2-ayatanaappindicator3-0.1
```
Command output gives:
```
2023-03-30 11:15:16: Initialising Email OAuth 2.0 Proxy (version 2023-03-28) from config file /home/wtcline/email-oauth2-proxy/emailproxy.config
2023-03-30 11:15:16: Starting IMAP server at localhost:1993 (unsecured) proxying outlook.office365.com:993 (SSL/TLS)
2023-03-30 11:15:16: Starting POP server at localhost:1995 (unsecured) proxying outlook.office365.com:995 (SSL/TLS)
```
Viewing bound interfaces with `netstat` shows:
```
Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name
tcp6       0      0 ::1:1993                :::*                    LISTEN      2784/python3
tcp6       0      0 ::1:1995                :::*                    LISTEN      2784/python3
```

Fedora 37:
```
$ sudo yum install gnome-shell-extension-appindicator python3-pip
$ pip3 install -r requirements.txt
<log out and back in>
$ gnome-extensions enable appindicatorsupport@rgcjonas.gmail.com
```
Command output gives:
```
2023-03-30 11:34:49: Initialising Email OAuth 2.0 Proxy (version 2023-03-28) from config file /home/wtcline/email-oauth2-proxy/emailproxy.config
2023-03-30 11:34:49: Starting IMAP server at localhost:1993 (unsecured) proxying outlook.office365.com:993 (SSL/TLS)
2023-03-30 11:34:49: Starting POP server at localhost:1995 (unsecured) proxying outlook.office365.com:995 (SSL/TLS)
```
Viewing bound interfaces with `netstat` shows:
```
Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name
tcp6       0      0 ::1:1995                :::*                    LISTEN      3916/python3
[...]

PR #153, opened 2023-03-30
Revert "Add IPv6 binding with dual-stack support (#140)"

wtcline-intc:
This reverts commit 966e6e23376afc08c4429e8d6b3b56e23b175565.

Issue #152 

This isn't an ideal solution, but it does fix the immediate issue.


simonrob:
See discussion in #152 – this has been resolved by improving the proxy's documentation.

PR #198, opened 2023-10-07
move fernet creation to helper method

terencehonles:
This change moves the fernet creation to a helper method in order to make it easier to either encrypt or decrypt values in the config/cache file. The motivation for this change is to make it easier to debug issues or to re-use OAuth tokens obtained outside the proxy.

### DIFF
--- a/emailproxy.py
+++ b/emailproxy.py
@@ -50,7 +50,7 @@ with warnings.catch_warnings():
     import asyncore
 
 # for encrypting/decrypting the locally-stored credentials
-from cryptography.fernet import Fernet, InvalidToken
+from cryptography.fernet import Fernet, InvalidToken, MultiFernet
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import hashes
 from cryptography.hazmat.primitives.kdf.pbkdf2 import PBKDF2HMAC
@@ -461,8 +461,8 @@ class AppConfig:
     _PARSER_LOCK = threading.Lock()
 
     # note: removing the unencrypted version of `client_secret_encrypted` is not automatic with --cache-store (see docs)
-    _CACHED_OPTION_KEYS = ['token_salt', 'access_token', 'access_token_expiry', 'refresh_token', 'last_activity',
-                           'client_secret_encrypted']
+    _CACHED_OPTION_KEYS = ['access_token', 'access_token_expiry', 'client_secret_encrypted', 'last_activity',
+                           'refresh_token', 'token_salt', 'token_iterations']
 
     # additional cache stores may be implemented by extending CacheStore and adding a prefix entry in this dict
     _EXTERNAL_CACHE_STORES = {'aws:': AWSSecretsManagerCacheStore}
@@ -570,34 +570,82 @@ class AppConfig:
             Log.error('Error saving state to cache store file at', cache_store_identifier, '- is the file writable?')
 
 
-class OAuth2Helper:
-    class TokenRefreshError(Exception):
-        pass
+class Cryptographer:
+    ITERATIONS = 870000  # taken from cryptography's suggestion of using Django's defaults
+    LEGACY_ITERATIONS = 100000
 
-    @staticmethod
-    def get_encryption_info(config, username, password):
-        """Returns a tuple of the fernet encrypter/decrypter and the random salt used in base64. This fernet is used to
-        store sensitive information for this account, including any stored tokens."""
-        token_salt = config.get(username, 'token_salt', fallback=None)
-        decoded_salt = None
+    def __init__(self, config, username, password):
+        """Creates a cryptographer which allows encrypting and decrypting sensitive information for this account,
+        including any stored tokens."""
+        self._salt = None
 
         # try to base64 decode the existing salt
+        token_salt = config.get(username, 'token_salt', fallback=None)
         if token_salt:
             try:
-                decoded_salt = base64.b64decode(token_salt.encode('ascii'))  # catch incorrect third-party proxy guide
+                self._salt = base64.b64decode(token_salt.encode('ascii'))  # catch incorrect third-party proxy guide
             except (binascii.Error, UnicodeError):
                 Log.info('%s: Invalid `token_salt` value found in config file entry for account %s - this value is not '
                          'intended to be manually created; generating new `token_salt`' % (APP_NAME, username))
 
         # generate a new salt if the salt cannot be decoded or this is the initial run
-        if not decoded_salt:
-            decoded_salt = os.urandom(16)
-            token_salt = base64.b64encode(decoded_salt).decode('ascii')
+        if not self._salt:
+            self._salt = os.urandom(16)
 
-        # generate encrypter/decrypter based on password and random salt
-        key_derivation_function = PBKDF2HMAC(algorithm=hashes.SHA256(), length=32, salt=decoded_salt, iterations=100000,
-                                             backend=default_backend())
-        return (Fernet(base64.urlsafe_b64encode(key_derivation_function.derive(password.encode('utf-8')))), token_salt)
+        # try to read the user configured token iterations
+        try:
+            iterations = int(config.get(username, 'token_iterations', fallback=self.LEGACY_ITERATIONS))
+        except ValueError:
+            iterations = self.LEGACY_ITERATIONS
+
+        # the first fernet is the primary fernet so sort the iterations count descending
+        self._iterations_options = sorted({self.ITERATIONS, iterations, self.LEGACY_ITERATIONS}, reverse=True)
+
+        # generate encrypter/decrypter based on the password and a random salt
+        password_bytes = password.encode('utf-8')
+        self._fernets = [
+            Fernet(base64.urlsafe_b64encode(
+                PBKDF2HMAC(algorithm=hashes.SHA256(), length=32, salt=self._salt, iterations=iterations,
+                           backend=default_backend()).derive(password_bytes)))
+            for iterations in self._iterations_options
+        ]
+        self.fernet = MultiFernet(self._fernets)
+
+    @property
+    def salt(self):
+        return base64.b64encode(self._salt).decode('ascii')
+
+    @property
+    def iterations(self):
+        return self._iterations_options[0]
+
+    def encrypt(self, value):
+        return self.fernet.encrypt(value.encode('utf-8')).decode('utf-8')
+
+    def decrypt(self, value):
+        return self.fernet.decrypt(value.encode('utf-8')).decode('utf-8')
+
+    def requires_rotation(self, value):
+        try:
+            # if the first fernet works, everything is up to date
+            self._fernets[0].decrypt(value.encode('utf-8'))
+            return False
+        except InvalidToken:
+            try:
+                # check to see if any fernet can decrypt the value
+                self.decrypt(value)
+            except InvalidToken:
+                return False
+
+            return True
+
+    def rotate(self, value):
+        return self.fernet.rotate(value.encode('utf-8')).decode('utf-8')
+
+
+class OAuth2Helper:
+    class TokenRefreshError(Exception):
+        pass
 
     @staticmethod
     def get_oauth2_credentials(username, password, reload_remote_accounts=True):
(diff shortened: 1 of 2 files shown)


emailproxy.py
```diff
@@ -574,6 +574,25 @@ class OAuth2Helper:
...
+        if not token_salt:
+            token_salt = base64.b64encode(os.urandom(16)).decode('utf-8')
+            config.set(username, 'token_salt', token_salt)
+
+        # generate encrypter/decrypter based on password and random salt
+        try:
+            decoded_salt = base64.b64decode(token_salt.encode('utf-8'))  # catch incorrect third-party proxy guide
+        except binascii.Error:
+            return ('%s: Invalid `token_salt` value found in config file entry for account %s - this value is not '
+                    'intended to be manually created; please remove and retry' % (APP_NAME, username))
```
terencehonles, line 591: This could also be rewritten to stomp on previous values and just show a warning instead.

simonrob: [This](https://hkdb.medium.com/office365-via-oauth2-with-geary-1c0a1429226e) is the post in question that for whatever reason gave that incorrect advice and led to a few invalid issues being opened (and this code being added). Traffic seems to have subsided from that source recently, so yes, replacing the value with a new salt and just showing a warning would probably be fine. So, something like:

```python
        # we hash locally-stored tokens with the given password
        token_salt = config.get(username, 'token_salt', fallback=None)

        # generate encrypter/decrypter based on password and random salt
        try:
            decoded_salt = base64.b64decode(token_salt.encode('utf-8'))  # catch incorrect third-party proxy guide
        except binascii.Error:
            Log.info('%s: Invalid `token_salt` value found in config file entry for account %s - this value is not '
                    'intended to be manually created; please remove and retry' % (APP_NAME, username))
            token_salt = None

        if not token_salt:
            token_salt = base64.b64encode(os.urandom(16)).decode('utf-8')

        config.set(username, 'token_salt', token_salt)
```

simonrob (changes requested):
Thanks for the PR. I don't have strong feelings either way about this, so it'd be useful if you could give a bit more detail about how it makes things easier. Is that because you'd rewrite or monkey-patch this function to change its behaviour? And how would it help with debugging issues or reusing externally-obtained OAuth tokens?

terencehonles (commented):
> Thanks for the PR. I don't have strong feelings either way about this, so it'd be useful if you could give a bit more detail about how it makes things easier. Is that because you'd rewrite or monkey-patch this function to change its behaviour? And how would it help with debugging issues or reusing externally-obtained OAuth tokens?

It just makes it easier since the logic is encapsulated in one place and it makes it easier (possible) to run something like:

```python
>>> import emailproxy, getpass
>>> config = emailproxy.AppConfig.get()
>>> username = input('Username: ')
Username: example@example.com
>>> cryptographer = emailproxy.OAuth2Helper.get_encryption_info(config, username, getpass.getpass())[0]
Password: 
>>> print("Refresh token: ", emailproxy.OAuth2Helper.decrypt(cryptographer, config.get(username, 'refresh_token')))
...
```

It's also possible to write new tokens following the same pattern as ^^^.

I wouldn't expect someone to monkey patch the function, but it's true that it does allow a way to potentially change the encryption method, though that would mean that `OAuth2Helper.encrypt` and `OAuth2Helper.decrypt` would also need to be updated and if making the encryption method be configurable then it probably makes sense to move everything to a class and create an instance that has a single interface.

emailproxy.py
```diff
@@ -574,6 +574,31 @@ class OAuth2Helper:
...
+                Log.info('%s: Invalid `token_salt` value found in config file entry for account %s - this value is not '
+                         'intended to be manually created; generating new `token_salt`' % (APP_NAME, username))
+
+        # generate a new salt if the salt cannot be decoded or this is the initial run
+        if not decoded_salt:
+            decoded_salt = os.urandom(16)
+            token_salt = base64.b64encode(decoded_salt).decode('ascii')
+
+        # generate encrypter/decrypter based on password and random salt
+        key_derivation_function = PBKDF2HMAC(algorithm=hashes.SHA256(), length=32, salt=decoded_salt, iterations=100000,
```
terencehonles, line 598: It looks like this is a bit low based on current recommendations. I believe this is not something that can be just updated and two different fernets probably need to be created and the old one can be used as a fallback when decrypting. A change like this is out of scope of this PR, and would require confirming a migration would be necessary.

emailproxy.py
```diff
@@ -710,7 +720,6 @@ def get_account_with_catch_all_fallback(option):
                 if username not in config.sections():
                     config.add_section(username)  # in catch-all mode the section may not yet exist
                     REQUEST_QUEUE.put(MENU_UPDATE)  # make sure the menu shows the newly-added account
-                config.set(username, 'token_salt', token_salt)
```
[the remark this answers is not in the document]
simonrob: I think a tuple is the best choice, because to fix the `KeyError` we'd need to create a new section, and although the config is only saved on a successful login, a later save (for example on a successful login with a different account) would include this new section. We could of course detect that and ignore/remove empty sections, but I think returning a tuple from the new function is the simplest approach here.

emailproxy.py
simonrob: Yes – this is something I looked at improving a while ago, but it was hard to find a good way to migrate. Since it's a locally-stored encrypted value I felt the risk was quite low, but if you can suggest a way to do this that isn't onerous from the user side I'm happy to look at it.

terencehonles: It looks like it might be relatively straightforward to do with a https://cryptography.io/en/latest/fernet/#cryptography.fernet.MultiFernet but you'd want to probably store the number of iterations used so that way you could run the migration and re-save the config when needed.

terencehonles:
The added commit should address https://github.com/simonrob/email-oauth2-proxy/pull/198#discussion_r1359926540, but I left it as a separate commit in case the change should be backed out from this PR.

requirements-no-gui.txt
```diff
@@ -1,7 +1,7 @@
 # this file contains the proxy's core dependencies beyond inbuilt python packages
 # note that to use this file instead of the default requirements.txt you *must* pass the `--no-gui` option when starting
 # the proxy - see the script's readme for further details
-cryptography
+cryptography>=2.2
```
terencehonles, line 4: This is quite old, but provides the `.rotate` method on the `MultiFernet`

emailproxy.py
```diff
@@ -570,6 +570,79 @@ def _save_cache(cache_store_identifier, output_config_parser):
             Log.error('Error saving state to cache store file at', cache_store_identifier, '- is the file writable?')
 
 
+class Cryptographer:
+    ITERATIONS = 870000  # taken from cryptography's suggestion of using Django's defaults
+    LEGACY_ITERATIONS = 100000
```
simonrob, line 575: What do you think about keeping a list of previous iteration counts given that the recommendation changes fairly often? (e.g., from 480,000 in December 2022 to 870,000 under a year later)

emailproxy.py
```diff
@@ -570,6 +570,79 @@ def _save_cache(cache_store_identifier, output_config_parser):
...
+    def __init__(self, config, username, password):
+        """Creates a cryptographer which allows encrypting and decrypting sensitive information for this account,
+        including any stored tokens."""
+        self._salt = None
+
+        # try to base64 decode the existing salt
+        token_salt = config.get(username, 'token_salt', fallback=None)
+        if token_salt:
+            try:
+                self._salt = base64.b64decode(token_salt.encode('ascii'))  # catch incorrect third-party proxy guide
```
simonrob, line 586: What's the reason for switching to ascii here?

emailproxy.py
```diff
@@ -570,6 +570,79 @@ def _save_cache(cache_store_identifier, output_config_parser):
...
+            self._salt = os.urandom(16)
+
+        # try to read the user configured token iterations
+        try:
+            iterations = int(config.get(username, 'token_iterations', fallback=self.LEGACY_ITERATIONS))
+        except ValueError:
+            iterations = self.LEGACY_ITERATIONS
+
+        # the first fernet is the primary fernet so sort the iterations count descending
+        self._iterations_options = sorted({self.ITERATIONS, iterations, self.LEGACY_ITERATIONS}, reverse=True)
```
simonrob, line 602: I'm curious – I'd always do `self.ITERATIONS + iterations + self.LEGACY_ITERATIONS` here. What's the reason for `{self.ITERATIONS, iterations, self.LEGACY_ITERATIONS}`?

simonrob (commented):
I was just about to take a look at iteration upgrading via MultiFernet, but saw this. Thanks for the great work!

emailproxy.py
terencehonles: Well this change stores the number of iterations in the config, so right now the options it tries are: the initial default 100_000, if there's a value in the config, and the new default 870_000, going forward if the config is updated it will store the value in the config, so it should really never go past that and the legacy value is just for historical reasons.

Ideally it's just trying [ITERATIONS, configured_iterations] and if the config is up to date that's actually just [ITERATIONS] when uniqued.

emailproxy.py
terencehonles: The base64 alphabet is always going to be ascii so if it doesn't encode then the decoding is not going to work. If for some reason (unlikely) the config returns unicode that can't be encoded as 'utf-8' this would have still broken before.

emailproxy.py
```diff
@@ -570,6 +570,79 @@ def _save_cache(cache_store_identifier, output_config_parser):
...
+            Fernet(base64.urlsafe_b64encode(
+                PBKDF2HMAC(algorithm=hashes.SHA256(), length=32, salt=self._salt, iterations=iterations,
+                           backend=default_backend()).derive(password_bytes)))
+            for iterations in self._iterations_options
+        ]
+        self.fernet = MultiFernet(self._fernets)
+
+    @property
+    def salt(self):
+        return base64.b64encode(self._salt).decode('ascii')
```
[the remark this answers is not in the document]
terencehonles: This one is guaranteed to always work since the base 64 alphabet is ascii

emailproxy.py
terencehonles: since it's a set it will drop out any duplicates since `iterations` _should_ be `self.ITERATIONS`

emailproxy.py
```diff
@@ -570,6 +570,79 @@ def _save_cache(cache_store_identifier, output_config_parser):
...
+        if not self._salt:
+            self._salt = os.urandom(16)
+
+        # try to read the user configured token iterations
+        try:
+            iterations = int(config.get(username, 'token_iterations', fallback=self.LEGACY_ITERATIONS))
+        except ValueError:
+            iterations = self.LEGACY_ITERATIONS
+
+        # the first fernet is the primary fernet so sort the iterations count descending
```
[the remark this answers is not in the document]
terencehonles: done

emailproxy.py
```diff
@@ -570,6 +570,80 @@ def _save_cache(cache_store_identifier, output_config_parser):
             Log.error('Error saving state to cache store file at', cache_store_identifier, '- is the file writable?')
 
 
+class Cryptographer:
+    ITERATIONS = 870000  # taken from cryptography's suggestion of using Django's defaults
+    LEGACY_ITERATIONS = 100000
```
terencehonles, line 575:
```suggestion
    LEGACY_ITERATIONS = 100000  # fallback used when the config file does not specify the iteration count
```

terencehonles:
One thing that may not be obvious is that you can actually set the token_iteration in the config to a value larger than what the script provides. As is, as long as the original iterations value is either `Cryptographer.ITERATIONS` or `Cryptographer.LEGACY_ITERATIONS` then they will be tried as fallbacks and the value in the config would be re-written since it would be the primary fernet.

This could support a previous iterations value by allowing the `Cryptographer` to have a fallback_iterations passed to its constructor, and one could rotate the config manually but even that could be moved to its own function to allow a user to easily force higher values and move between them.

emailproxy.py
simonrob: Sorry, I had in my head that these values were lists after making the first comment above (and hence this would not work). Ignore this one!

simonrob:
I edited a few comments and removed an unnecessary `ValueError` check (if fallback is specified this doesn't happen). I think this is good to go?

simonrob (approved):

PR #201, opened 2023-10-16
allow importing as a script without GUI requirements installed

terencehonles:
This change updates the script initialization to not fail on import errors, to finish without checking if `--no-gui` or `--external-auth` has been passed in the command-line, and instead raises an error when the application is initializing if running in GUI mode and the requirements are not met.

Minor changes also included:
- The `--no-gui` option is now stored as `gui` and defaults to `True` which changes comparisons to not be double negatives but retains the existing behavior.
- The App instance is allowed a single argument which is the command-line to parse, which allows starting the app with arguments from an import.

### DIFF
--- a/emailproxy.py
+++ b/emailproxy.py
@@ -3130,6 +3130,10 @@ class App:
                 proxy.stop()
 
         if icon:
+            # work around a pystray issue with removing the macOS status bar icon when started from a parent script
+            if sys.platform == 'darwin':
+                # noinspection PyProtectedMember
+                icon._status_item.button().setImage_(None)
             icon.stop()
 
         # for the 'Start at login' option we need a callback to restart the script the first time this preference is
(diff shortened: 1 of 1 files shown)


terencehonles:
This will also allow packaging without all the requirements and this check https://github.com/simonrob/email-oauth2-proxy/pull/199/files#diff-8068eda2feddbc61595ced559aa73f8502813cf5cb1190a21e111ab1e8d19e88R71

simonrob (commented):
One benefit of the previous approach was that failed imports were more visible because they printed an error. With this new approach it's a lot harder to see which import is missing (and with #199 merged there won't necessarily be an easily-available requirements.txt to fix this).

At the moment, apart from for pystray and AppKit, is there any need to separate all of the imports and catch errors independently? If any single one fails and `--no-gui` isn't specified then we're going to exit anyway. So, perhaps a better approach is to bring back the original ArgumentParser, but use it to decide whether to display an error or not (i.e., if in GUI mode and imports fail).

PR #205, opened 2023-11-14
Fix trailing whitespace

brianjmurrell:


brianjmurrell:
Hey @simonrob.  Not sure if the whitespace on the empty lines is required in this file's format or not.  It just ends up being annoying in editors that are automagically trimming trailing whitespace for you as it creates a whitespace diff from upstream in your local working copy.

If it's required, obviously it's something downstreams are going to have to live with if they have automagic whitespace trimming.

simonrob:
Thanks - yes, this is not intended. I appreciate the fix!

simonrob (approved):

PR #208, opened 2023-11-27
Improve Python packaging

stumpylog:
Instead of spreading across 3 files, the dependecies can be combined into the pyproject.toml, instead of being dynamic.  Installation is basically the same, including the optional extra named `[gui]`.  For the Git clone installation, it's actually easier, as there is no need to specify one or 2 requirement files.


README.md
```diff
@@ -28,7 +28,7 @@ Begin by downloading the proxy via one of the following methods:
...
   <li><b>Install from <a href="https://pypi.org/project/emailproxy/">PyPI</a></b>: set up using <code>python -m pip install emailproxy[gui]</code>, download the <a href="https://github.com/simonrob/email-oauth2-proxy/raw/main/emailproxy.config">sample <code>emailproxy.config</code> file</a>, then <code>python -m emailproxy</code> to run; <i>or</i>,</li>
-  <li><b>Clone or <a href="https://github.com/simonrob/email-oauth2-proxy/archive/refs/heads/main.zip">download</a></b> (and star :-) the <a href="https://github.com/simonrob/email-oauth2-proxy/">GitHub repository</a>, then: <code>python -m pip install -r requirements-core.txt -r requirements-gui.txt</code> to install requirements, and <code>python emailproxy.py</code> to run.</li>
+  <li><b>Clone or <a href="https://github.com/simonrob/email-oauth2-proxy/archive/refs/heads/main.zip">download</a></b> (and star :-) the <a href="https://github.com/simonrob/email-oauth2-proxy/">GitHub repository</a>, then: <code>python -m pip .</code> or <code>python -m pip .[gui]</code>to install requirements (for headless or GUI respectively), and <code>python emailproxy.py</code> to run.</li>
```
simonrob, line 31: This is incorrect (missing `install` from the installation command, and running would be `python -m emailproxy`)

simonrob (commented):
Thanks for this suggestion.

Retention of the requirements.txt files during the move to pyproject.toml was actually intentional because of the [`plugins` version of the proxy](https://github.com/simonrob/email-oauth2-proxy/tree/plugins), which relies on a set of additional local files to enable its extra capabilities. While it works as-is when packaged if you are careful about paths, it's easy to go wrong if running from, say, a different directory. With the non-packaged approach (i.e., via requirements.txt) this far less likely to be an issue. This approach is also easier for anyone working on developing the proxy because there is no packaging layer to worry about.

Is there any particular benefit from removing these files that I'm missing?

PR #247, opened 2024-05-06
Implements certificate credential authentication using JWT assertions

jannispl:
Since Microsoft is no longer issuing client secrets with indefinite validity for O365 application registrations, I have decided to implement certificate credential authentication using JWT assertions. At this time, there seem to be no restrictions on the validity of self-signed certificates (I managed to upload and use a 10yr certificate to Entra just fine).

Microsoft even [recommends](https://learn.microsoft.com/entra/identity-platform/v2-oauth2-auth-code-flow#redeem-a-code-for-an-access-token) certificate authentication:
>For best security, we recommend using certificate credentials.

The workflow and technicals for certificate authentication are described [here](https://learn.microsoft.com/entra/identity-platform/certificate-credentials).

This PR adds two account configuration options `jwt_certificate_path` and `jwt_key_path`. These options are to be used in place of `client_secret` (i.e., that value should be removed). `jwt_certificate_path` is expected to be a path to a x509 pem-encoded certificate; `jwt_key_path` is expected to be a path to an unencrypted pem-encoded private key.

To generate the certificate and private key, the following command can be used:
```
openssl req -x509 -newkey rsa:4096 -keyout key.pem -out certificate.pem -sha256 -days 3650 -nodes -subj "/CN=email-oauth2-proxy"
```
The `certificate.pem` that is generated can then be uploaded to the application registration in the Entra admin console.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,13 +69,13 @@ In all cases, when registering, make sure your client is set up to use an OAuth 
 It is also highly recommended to use a scope that will grant "offline" access (i.e., a way to [refresh the OAuth 2.0 authentication token](https://oauth.net/2/refresh-tokens/) without user intervention).
 The [sample configuration file](https://github.com/simonrob/email-oauth2-proxy/blob/main/emailproxy.config) provides example scope values for several common providers.
 
-- Office 365: register a new [Microsoft identity application](https://learn.microsoft.com/en-us/entra/identity-platform/quickstart-register-app)
+- Office 365: register a new [Microsoft identity application](https://learn.microsoft.com/entra/identity-platform/quickstart-register-app)
 - Gmail / Google Workspace: register a [Google API desktop app client](https://developers.google.com/identity/protocols/oauth2/native-app)
 - AOL and Yahoo Mail (and subproviders such as AT&T) are not currently allowing new client registrations with the OAuth email scope – the only option here is to reuse the credentials from an existing client that does have this permission.
 
 The proxy supports [Google Cloud service accounts](https://cloud.google.com/iam/docs/service-account-overview) for access to Google Workspace Gmail.
-It also supports the [client credentials grant (CCG)](https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth2-client-creds-grant-flow) and [resource owner password credentials grant (ROPCG)](https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth-ropc) OAuth 2.0 flows.
-Please note that currently only Office 365 is known to support the CCG and ROPCG methods.
+It also supports the [client credentials grant (CCG)](https://learn.microsoft.com/entra/identity-platform/v2-oauth2-client-creds-grant-flow) and [resource owner password credentials grant (ROPCG)](https://learn.microsoft.com/entra/identity-platform/v2-oauth-ropc) OAuth 2.0 flows, and [certificate credentials (JWT)](https://learn.microsoft.com/entra/identity-platform/certificate-credentials).
+Please note that currently only Office 365 is known to support the CCG, ROPCG and certificate credentials methods.
 See the [sample configuration file](https://github.com/simonrob/email-oauth2-proxy/blob/main/emailproxy.config) for further details.
 
 
--- a/emailproxy.config
+++ b/emailproxy.config
@@ -123,6 +123,10 @@ documentation = Accounts are specified using your email address as the section h
 	- It is possible to create Office 365 clients that do not require a secret to be sent. If this is the case for your
 	setup, delete the `client_secret` line from your account's configuration entry (do not leave the default value).
 
+	- To use O365 certificate credentials instead of a client secret, delete the `client_secret` line and instead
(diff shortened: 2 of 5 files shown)


simonrob:
Thanks for the contribution!

In order to test before merging this I'll need access to an account that can be authenticated via JWT, which I don't have at the moment. Is this something you could provide for this purpose?

jannispl:
I could provide you with a temporary test account. How can I contact you privately?

simonrob:
That would be perfect. You can reach me via my GitHub username at Gmail.

jannispl:
I have sent you an email.

simonrob:
Thanks for providing the test account – I've just taken a look. I made a few minor edits:
- Remove the edit to `pyproject.toml` as this is not needed (but clarify the message here for future reference)
- Lower the minimum `pyjwt` version to 2.4 so that Python 3.6 is still supported
- Variable renaming (including in the configuration file) for consistency with the rest of the proxy
- Lint reformatting to match the project style
- Remove some of the JWT parameters that RFC 7519 specifies as optional and O365 does not require
- Add basic documentation

Just so there is a record, here's a sample configuration file entry using this method:

```ini
[your-email-address@example.com]
permission_url = https://login.microsoftonline.com/*** your tenant id here ***/oauth2/v2.0/authorize
token_url = https://login.microsoftonline.com/*** your tenant id here ***/oauth2/v2.0/token
oauth2_scope = https://outlook.office365.com/IMAP.AccessAsUser.All https://outlook.office365.com/SMTP.Send offline_access
redirect_uri = http://localhost
client_id = *** your client id here ***
jwt_certificate_path = /path/to/certificate.pem
jwt_key_path = /path/to/key.pem
```

Unless there are any other features to be added, I think this is probably ready to be merged. Let me know what you think?

jannispl:
Thanks for taking your time to polish this - it's looking really good now.

Re:
>- Remove some of the JWT parameters that RFC 7519 specifies as optional and O365 does not require

They might not be required by the spec, but at least for `jti` [Microsoft's documentation](https://learn.microsoft.com/en-us/entra/identity-platform/certificate-credentials) is worded a little more strict:

>The identifier value **must** be assigned in a manner that ensures that there's a negligible probability that the same value will be accidentally assigned to a different data object [...]

I also found [this page](https://learn.microsoft.com/en-us/entra/msal/dotnet/acquiring-tokens/web-apps-apis/confidential-client-assertions) which states:

>The [claims expected by Microsoft Entra ID](https://learn.microsoft.com/en-us/azure/active-directory/develop/certificate-credentials) in the signed assertion are: [...]

and

>If you want to provide your own claims, including the **mandatory** claims expected by Microsoft Entra ID, pass in false for the mergeWithDefaultClaims parameter.

simonrob:
Interesting – thanks for following up. The original page isn't completely clear – it essentially just lists the properties defined in the RFC. Because of this, I determined through experimentation which ones O365 does and does not actually require for OAuth 2.0 login. I hadn't seen the other link you've found, but in its [Crafting the assertion section](https://learn.microsoft.com/en-us/entra/msal/dotnet/acquiring-tokens/web-apps-apis/confidential-client-assertions#crafting-the-assertion) it gives this example:

```c#
// no need to add exp, nbf as JsonWebTokenHandler will add them by default.
var claims = new Dictionary<string, object>()
{
    { "aud", tokenEndpoint },
    { "iss", clientId },
    { "jti", Guid.NewGuid().ToString() },
    { "sub", clientId }
};
```

As explained in the comment, the `exp` property is added automatically by `JsonWebTokenHandler` in this example. The proxy currently sets a 5-minute expiry time here. Interestingly, Microsoft's default expiry time is [60 minutes](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/9adc0ef59efd7f2151f6b783608919fd0e8f5226/src/Microsoft.IdentityModel.Tokens/TokenHandler.cs#L26), though their documentation suggests using a value of [5-10 minutes at most](https://learn.microsoft.com/en-us/entra/msal/dotnet/acquiring-tokens/web-apps-apis/confidential-client-assertions#client-assertions). It's safe to stick to the proxy's current approach for this property, I think.

From testing, I found that the `nbf` property is not actually required by O365, though I agree that we should probably add it back since it seems like it may be expected, and the default that O365 sets this value to (the Unix epoch) does not seem sensible. The `iat` property is not mentioned on this page, and not currently added in [JsonWebTokenHandler](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs) either (though that [hasn't always been the case](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/a912c48eaac792650dc414b9c7e12cac63ae25de/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs#L165)).

So: long story short, I think `nbf` and `jti` should be added, but we can leave `iat` out. What do you think?

jannispl:
>So: long story short, I think `nbf` and `jti` should be added, but we can leave `iat` out. What do you think?

Sounds reasonable to me.

PR #254, opened 2024-06-01
Minor documentation fix

wtcline-intc:
The version check issue with pywebview is in Ubuntu 24.04 (Noble), not 22.04 (Jammy).


wtcline-intc:
I've made this typo countless times already, too...

simonrob:
Thanks for pointing this out. I couldn't merge it directly because the repository has signed commits enabled, but I've made the change.

PR #278, opened 2024-09-05
Add more or less standard header for the Python3 script

ashevche:
The script is using Python 3 interpreter and also it's good to have internal encoding defined. Taking all these into account, add a more or less standard header to the script.

### DIFF
--- a/emailproxy.py
+++ b/emailproxy.py
@@ -1,4 +1,6 @@
 #!/usr/bin/env python3
+# -*- coding: UTF-8 -*-
+# vim: ts=4 sw=4 et ai si
 
 """A simple IMAP/POP/SMTP proxy that intercepts authenticate and login commands, transparently replacing them with OAuth
 2.0 authentication. Designed for apps/clients that don't support OAuth 2.0 but need to connect to modern servers."""


ashevche:
Now it's signed. I dunno what GitHub wants from me. Probably a bug in GitHub.

simonrob:
Please could you explain why you think these are necessary? `UTF-8` is the default for Python 3, and `vim` is one editor amongst many others. I don't currently see the need for these, so it would have to be a pretty strong argument to add them.

---

> Now it's signed. I dunno what GitHub wants from me. Probably a bug in GitHub.

This is not a bug; the repository has a rule to require signed commits.

simonrob:
Without this change _vim_ may break PEP-8, but that's just one editor, and it's not the one I use to maintain this project. As a result I really don't see any particular reason to add it – I'm sure all sorts of other editors have their own configuration comments like this, but that's not a concern if people who use them aren't actually maintainers of the project.

As the Python 3 documentation clearly explains, [`If no encoding declaration is found, the default encoding is UTF-8`](https://docs.python.org/3/reference/lexical_analysis.html#encoding-declarations).

---

> Why the commit is still in _Unverified_ state?

The error message explains this (emphasis mine): "The base branch requires _all_ commits to be signed".

PR #302, opened 2024-10-25
Add support for OAUTH2 device flow (RFC 8628)

a-ilin:
Hi! Thank you for the great tool!

Could you please take a look at this PR?

## Description

This PR adds support for OAUTH2 device flow:
- RFC: https://datatracker.ietf.org/doc/html/rfc8628
- Microsoft: https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth2-device-code

Device flow is particularly useful on headless systems, as it does not require usage of the redirect URL.

I successfully tested this PR with a free Microsoft account (`outlook.com`), and own application `client_id`.
The application (`client_id`) must be explicitly allowed to use the device flow in the Azure console.

## Implementation

The following properties are introduced into the request propagated from OAUTH2 to the main app:
- `need_response`: A boolean specifying the flow implies a response from a user to be provided to the proxy in order to retrieve the tokens. It is `false` for device flow, and `true` for others.
- `user_code`: A (short) string, which should be provided to the user for authentication at the service provider. User should manually enter it at the provider's web page.

The following changes are done to the UI, both graphical and console:
- external auth: the user code is printed to the console together with the verification URI.
- local server auth: the user code is printed to the console together with the verification URI.
- icon: the user code is added to the title of the authorization window.

## QA

I tested the proxy on Linux systems, in container and on desktop.
The following test cases are successful:
- external auth: the verification URI and user code are printed to the console.
- local server auth: the verification URI and user code are printed to the console.

**Attention!** On my systems I cannot make GUI work at all (even without my changes). The icon in the tray is always frozen, without notifications, and without context menu.
Therefore I cannot validate that the changes are correct for GUI interaction.

### DIFF
--- a/emailproxy.config
+++ b/emailproxy.config
@@ -155,6 +155,10 @@ documentation = Accounts are specified using your email address as the section h
 		attempts before the first valid login, pre-encrypting account entries is highly recommended. See the example
 		script at https://github.com/simonrob/email-oauth2-proxy/issues/61#issuecomment-1259110336.
 
+	- The proxy supports the device OAUTH2 flow (RFC 8628), which may better suit headless systems. With this flow
+	the proxy	receives authorization response directly from the service provider, therefore `redirect_uri` is not needed.
+	To use this flow set `oauth2_flow = device`.
+
 	Gmail customisation:
 	- The proxy supports the use of service accounts with Gmail for Google Workspace (note: normal Gmail accounts do not
 	support this method). To use this option, add an account entry as normal, but do not add a `permission_url` value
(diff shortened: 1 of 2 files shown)


simonrob:
Thanks for the work adding this feature. I'll need time to review it - hopefully in the next few weeks.

simonrob:
I had a look at this just now. I had to adjust your approach a little in order to maintain GUI compatibility. I also tried to stick to the current architecture of the proxy more closely where possible.

Because the commits in your branch aren't signed (so can't be merged), I made a new branch: [device-authorisation-grant](https://github.com/simonrob/email-oauth2-proxy/tree/device-authorisation-grant). Could you take a look?

a-ilin:
Thank you for the rework! I had a look over the changed implementation, and verified the case with `--no-gui` and `--external-auth` options. The proxy works fine!

The only issue I noticed in the code is that the `interval` field within the authorization response should be optional, and in case of absence it should be set to 5 seconds, as per the RFC. Currently there is no fallback in `get_oauth2_authorisation_tokens`.

simonrob:
Thanks for taking a look, and for spotting the optionality of the interval – ab0fbd300632f430962759f4c9970a57d55e406f re-adds this. I'll go ahead and merge this new feature now.

simonrob:
Closing this PR as its features have been added in b62dd5843904aa9cac804d0a3e42cbcb97584de4. Thanks @a-ilin for the work on this.

PR #304, opened 2024-11-16
Document WebKit renderer work arounds

wtcline-intc:
Sometimes PyWebKit doesn't always render as expected.  Document a few known workarounds for rendering issues.


simonrob:
Thanks for the contribution. I presume this is a GTK-specific issue, but could you confirm?

wtcline-intc:
> Thanks for the contribution. I presume this is a GTK-specific issue, but could you confirm?

I'm not sure.  I ran into this issue on a GTK-based system, but I would hypothesize that disabling GL-based rendering would be an effective workaround on any graphics engine when hardware-based rendering fails.

Issue #265, opened 2024-07-27
Emailproxy.exe is being reported as a containing a virus by McAffe antivirus.


Rick-G1:
Because of the common problem of many PyPI compilations being detected as a virus, such as Emailproxy.exe is, would it be possible to list the MD5 hash for at least the Windows version of the executable file emailproxy.exe.   This would give people downloading the package a degree of confidence that the downloaded files do not contain a virus.

This can be done in Windows with "certutil -hashfile  emailproxy.exe MD5"  and then publish the hash number.  (The number I have on the latest one I've downloaded [2024_05_25]  is   65dfebe8463511e0f8a69f6cc4f49b8f)

Thanks!!

simonrob:
Thanks for the suggestion. It's a shame that virus scanners incorrectly do this, but always worth reporting these incorrect results to them to improve detection (different providers support this in different ways, but here's the [AVG reporting page](https://secure.avg.com/submit-sample), for example).

Re: MD5 hash – I'm not sure what this would achieve, given that the compiled version of the proxy is built [entirely automatically](https://github.com/simonrob/email-oauth2-proxy/blob/build/.github/workflows/pyinstaller.yml), and any person or process that has access to set edit the release page to add the hash could also maliciously edit the compiled file. Still, in case it helps, the hash from my side is also `65dfebe8463511e0f8a69f6cc4f49b8f`.

Rick-G1:
Hi: 

 

First of all --- Thank you so much for your assistance !!  I really appreciate it !!  Thanks for reporting the hash value.

 

Tried to report this file to McAffe, but their reporting mechanism said that the 32MB file was "too big"…  I also found a rather interesting web site that checks about 75 different antivirus tools.  The site is  https://www.virustotal.com/gui/home/upload and you just drag the suspected file onto it.   It shows that emailproxy.exe shows problems on 14 of 75 tools.   

 

As for the hash…. If you change a single line of code (either in your python source or after compilation) the MD4 hash will be different, and you'd have to be very creative to then add "stuff" to malicious code to make the number the same.   So this gives people confidence in the fact that the data supplied is what the author intended. 

 

[...]

PR #330, opened 2025-02-15
Added Nuitka artifacts via gh-actions

sommerf-lf:
As our company's AV did always flag the pyinstaller version (as also noted in releases text and several discussions) and it doesn't quite fit our environment to run venvs barebone, I decided to add Nuitka compilation. ([Nuitka](https://nuitka.net/) converts python code to C/C++ and then compiles it, which makes the release standalone.)
I'd appreciate it, if you would give me feedback and if possible merge my changes into the main repo. I was not quite sure what the relationship between main and build is, but as the workflow files are in build, I worked on this branch. (It should be able to be merged into main without any conflicts.)

My changes:
- setup of nuitka.yml workflow
- modification of emailproxy.py to ignore version checks when running as Nuitka
- tested Nuitka artifact (only windows testing was possible, but should also work on macos; I tested the default behaviour used in our use case)

see for artifacts: https://github.com/sommerf-lf/email-oauth2-proxy/releases/tag/build-1fcb1933
- [emailproxy-2024-11-11_nuitka-macOS.zip](https://github.com/sommerf-lf/email-oauth2-proxy/releases/download/build-1fcb1933/emailproxy-2024-11-11_nuitka-macOS.zip)

- [emailproxy-2024-11-11_nuitka-Windows.zip](https://github.com/sommerf-lf/email-oauth2-proxy/releases/download/build-1fcb1933/emailproxy-2024-11-11_nuitka-Windows.zip)

image

### DIFF
--- a/.github/workflows/nuitka.yml
+++ b/.github/workflows/nuitka.yml
@@ -1,0 +1,76 @@
+name: Add Nuitka assets to latest release
+
+on:
+  push:
+
+permissions:
+  contents: write
+
+jobs:
+  get_tag:
+    runs-on: ubuntu-latest
+    outputs:
+      output: ${{ steps.latest_tag.outputs.tag }}
+
+    steps:
+      - uses: actions/checkout@v4
+      - id: latest_tag
+        run: |
+          git fetch -a
+          echo "tag=$(git tag --sort -committerdate | head -n 1)" >> $GITHUB_OUTPUT
+
+  build:
+    needs: get_tag
+
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [macos-latest, windows-latest]
+
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.x'
+
+      - run: |
+          python -m pip install --upgrade pip
+          python -m pip install wheel
+          python -m pip install -r requirements-core.txt -r requirements-gui.txt nuitka
+          python -m pip install certifi
+          python -m pip install zstandard ordered-set
+          python -m pip install boto3 requests google-auth
+          python -m pip install imageio
+
+      - if: runner.os == 'macOS'
+        run: |
+            python -m nuitka --standalone --onefile --include-plugin-directory=. --include-package=timeago.locales.en_short --include-package=certifi --include-package=boto3 --include-package=requests --include-package=google --macos-create-app-bundle --macos-app-icon=icon.png --assume-yes-for-downloads --output-dir=dist emailproxy.py
+            rm -rf dist/emailproxy.build dist/emailproxy.dist dist/emailproxy.onefile-build
+      - if: runner.os == 'Windows'
+        run: |
+            python -m nuitka --standalone --onefile --include-plugin-directory=. --include-package=timeago.locales.en_short --include-package=certifi --include-package=boto3 --include-package=requests --include-package=google --windows-icon-from-ico=icon.png --assume-yes-for-downloads  --output-dir=dist emailproxy.py
+            Remove-Item -Path "dist/emailproxy.build","dist/emailproxy.dist","dist/emailproxy.onefile-build" -Recurse -Force
+
+      - run: |
+          move LICENSE . 2> nul || { shopt -s expand_aliases; alias move=mv; }
+          move LICENSE dist/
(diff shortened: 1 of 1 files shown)


simonrob:
Thanks for the PR. It's a simple enough thing to switch to a different packaging system, but I originally chose PyInstaller as it seemed to be the most popular choice in other projects. This popularity of course can lead to issues (e.g., the virus scanner false positives), but since I don't use the pre-built versions myself I didn't really see this as a major problem.

As it happens, I did some work a while ago to make the proxy compatible with Nuitka builds, and the instructions are already [in the readme](https://github.com/simonrob/email-oauth2-proxy?tab=readme-ov-file#advanced-configuration). Can you explain why the version check change is needed? I didn't run into any issues here in my own testing.

sommerf-lf:
You're absolutely right! It's just that many users might be discouraged if they need to compile things themselves. Also it's very very unfortunate that many AI-based antivirus programs tend to flag PyInstaller by default. 

I'm planning to review the version checking again, probably next week when I have a bit more time. You might be right that the changes aren't necessary anymore. A few months ago, when I was using Nuitka, it had issues with these checks. I noticed you've since switched from `pkg_resources` to `packaging`, which might have addressed the earlier problems. I'll look into this and keep you updated on what I find.

Thanks a lot for the excellent tool, by the way!

But would you generally be open to the idea of publishing the Nuitka release as well?

simonrob:
I'm happy to switch if it turns out Nuitka packaging is preferable. Could you and/or any other commenters confirm that this option definitely doesn't trigger virus warnings?

At the moment, rather than merging this PR, I'd probably just make the changes in the existing build script to use Nuitka instead since I'm not convinced the other changes are needed.

In addition, if PyInstaller is causing significant problems, as noted above, you can already [build this yourself](https://github.com/simonrob/email-oauth2-proxy?tab=readme-ov-file#advanced-configuration) – this has been possible since 50d07c904b2de16349d4cc267321092cff9467da (July 2024) which made changes specifically to support Nuitka.

simonrob:
Thanks for the detailed summary. Like I said, I'm happy to switch if it's a better option. I don't think I'll keep both Nuitka _and_ PyInstaller though - I can imagine this leading to questions about which one to use, and potential interoperability issues. There are also PyInstaller-specific things in the proxy at the moment that I'd be happy to remove.

Let me know about the other changes and I'll take a look when I next get chance.

sommerf-lf:
I had a brief look again at what I did when I earlier (April 2023; on commit 779e70e557d0e66fe583acaeb1cd3fdccba5fb92) made a quickly thrown together way to run using Nuitka. I had to remove the usages of `pkg_resources`. But this is not necessary anymore, since Nutika seems to play nicely with `packaging`. (so I removed the 2nd unnecessary python change to check if running Nuitka)
To not further complicate this, I still suggest to offer both, with a sentence of something like the following

> Please note that the pre-built executables provided here are packaged automatically directly from the source using [PyInstaller](https://github.com/pyinstaller/pyinstaller) and alternatively [Nuitka](https://nuitka.net/). It is possible that virus scanners will raise false positive warnings about these files. These false positives are common with packaged/compiled python distributables. The PyInstaller and Nuitka builds have the same features and just have different tools for packaging. Pick whichever you AV seems to favor (Windows Defender seems to dislike PyInstaller; For more information look at this [discussion](https://github.com/simonrob/email-oauth2-proxy/pull/330)). Other significantly relevant differences are not known of here. 
If it helps you are welcome to report this problem to the various virus scanning providers (e.g., see https://github.com/simonrob/email-oauth2-proxy/issues/265).

simonrob:
I looked at this in more detail in preparation for a new release, in the hope that, as discussed above, although Nuitka is actually _more_ likely to be detected as malicious, it seems to be detected by a different set of scanners, so may offer respite to those struggling with the PyInstaller false positive issues.

I couldn't get the GitHub-built macOS release to work via the approach in this PR (even though it has built fine for me in the past with my own local macOS configuration). In the end, I decided to add Nuitka support to the existing build file, and switched to the [official Nuitka GitHub Action](https://github.com/Nuitka/Nuitka-Action). As a result, the [2025-03-14 release](https://github.com/simonrob/email-oauth2-proxy/releases/tag/2025-03-14) now has both PyInstaller and Nuitka assets.

I'll close this PR as it is no-longer necessary, but thanks again for the suggestion and investigation here.

PR #345, opened 2025-03-31
Create Dockerfile

tsgoff:
init
```
docker run   \
    -p 127.0.0.1:1587:1587 \
    -ti -v /opt/oauth2-proxy/emailproxy.config:/app/emailproxy.config \
    bios/email-oauth2-proxy python -m emailproxy --external-auth --no-gui
```

daemon:
```
docker run -d --name emailproxy \
            --restart="always" \
           -p 127.0.0.1:1993:1993 \
           -p 127.0.0.1:1587:1587 \
           -v /opt/oauth2-proxy/emailproxy.config:/app/emailproxy.config \
           bios/email-oauth2-proxy
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,23 @@
+FROM python:3.11-slim AS builder
+
+WORKDIR /app
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    gcc \
+    libffi-dev \
+    python3-dev \
+  && rm -rf /var/lib/apt/lists/*
+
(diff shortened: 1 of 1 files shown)


simonrob:
This has previous been proposed and not merged (see #100). I don't think the situation has changed, and in fact your suggestion is probably less flexible than the one in the earlier PR as it specifies options within the Dockerfile that may not apply to all setups.

The readme links to an [example Docker configuration](https://github.com/blacktirion/email-oauth2-proxy-docker), which is the preference for now.

Issue #237, opened 2024-03-13
TypeError("unsupported operand type(s) for +: 'int' and 'str'")


FlashIvano:
```yaml
version: '3'

services:      
  email-oauth2-proxy:
    image: blacktirion/email-oauth2-proxy-docker:2024.02.22

    container_name: email-oauth2-proxy
    network_mode: 'host'
    volumes:
      - /opt/compose/email-oauth2-proxy/cache:/cache
      - /opt/compose/email-oauth2-proxy/conf:/config
      - /etc/localtime:/etc/localtime:ro
    restart: unless-stopped
    environment:
      LOGFILE: true
      DEBUG: false
      CACHE_STORE: /cache/credstore.config
      #LOCAL_SERVER_AUTH: true #Optional

```

cat conf/emailproxy.config

```ini
[IMAP-143]
server_address = partner.outlook.cn
server_port = 993

[POP-110]
server_address = partner.outlook.cn
server_port = 995

[SMTP-25]
server_address = smtp.partner.outlook.cn
server_port = 587
starttls = True

[AUSER@MYOFFICE365CN.DOMAIN]
token_url = https://login.partner.microsoftonline.cn/0000000-0000-0000-0000-0000000/oauth2/token
oauth2_scope = https://partner.outlook.cn/.default
oauth2_flow = client_credentials
redirect_uri = http://localhost
client_id = 111111-1111-1111-1111-111111
client_secret = <secret_here>

[emailproxy]
delete_account_token_on_password_error = True
encrypt_client_secret_on_first_use = False
allow_catch_all_accounts = False
```

docker compose log:

```
email-oauth2-proxy  | 2024-03-13 17:16:00,749: Starting IMAP server at [::]:143 (unsecured) proxying partner.outlook.cn:993 (SSL/TLS)
email-oauth2-proxy  | 2024-03-13 17:16:00,751: Starting POP server at [::]:110 (unsecured) proxying partner.outlook.cn:995 (SSL/TLS)
email-oauth2-proxy  | 2024-03-13 17:16:00,751: Starting SMTP server at [::]:25 (unsecured) proxying smtp.partner.outlook.cn:587 (STARTTLS)
email-oauth2-proxy  | 2024-03-13 17:16:00,751: Initialised Email OAuth 2.0 Proxy - listening for authentication requests. Connect your email client to begin
[...]

PR #351, opened 2025-04-25
Support resource field for legacy ROPC flow; normalize expires_in handling

Zhehan-Z:
This pull request enhances the OAuth 2.0 authentication flow in `emailproxy.py` by introducing support for `oauth2_resource` as an alternative to `oauth2_scope`, refining the handling of different OAuth 2.0 flows, and improving error handling and token normalization. Below are the key changes grouped by theme:

### Support for `oauth2_resource`:
* Added `oauth2_resource` as an optional configuration parameter retrieved via `AppConfig.get_option_with_catch_all_fallback`. This allows specifying a resource instead of a scope for OAuth 2.0 flows. (`[emailproxy.pyR742](diffhunk://#diff-8068eda2feddbc61595ced559aa73f8502813cf5cb1190a21e111ab1e8d19e88R742)`)
* Updated the validation logic to require either `oauth2_scope` or `oauth2_resource` when `token_url` and `client_id` are provided. (`[emailproxy.pyL755-R756](diffhunk://#diff-8068eda2feddbc61595ced559aa73f8502813cf5cb1190a21e111ab1e8d19e88L755-R756)`)
* Modified the `get_oauth2_authorisation_tokens` method to include `oauth2_resource` in the request parameters when applicable. (`[[1]](diffhunk://#diff-8068eda2feddbc61595ced559aa73f8502813cf5cb1190a21e111ab1e8d19e88L1169-R1176)`, `[[2]](diffhunk://#diff-8068eda2feddbc61595ced559aa73f8502813cf5cb1190a21e111ab1e8d19e88L1195-R1237)`)

### Refinements to OAuth 2.0 flow handling:
* Adjusted logic to set the default flow to `authorization_code` unless the flow is explicitly specified as `device` or `password`. (` [emailproxy.pyR887](diffhunk://#diff-8068eda2feddbc61595ced559aa73f8502813cf5cb1190a21e111ab1e8d19e88R887) `)

### Improved error handling and token normalization:
* Enhanced the token retrieval process to normalize the `expires_in` value to an integer, with error logging for invalid values. (`[emailproxy.pyL1195-R1237](diffhunk://#diff-8068eda2feddbc61595ced559aa73f8502813cf5cb1190a21e111ab1e8d19e88L1195-R1237)`)

### DIFF
--- a/emailproxy.py
+++ b/emailproxy.py
@@ -739,6 +739,7 @@ class OAuth2Helper:
         permission_url = AppConfig.get_option_with_catch_all_fallback(config, username, 'permission_url')
         token_url = AppConfig.get_option_with_catch_all_fallback(config, username, 'token_url')
         oauth2_scope = AppConfig.get_option_with_catch_all_fallback(config, username, 'oauth2_scope')
+        oauth2_resource = AppConfig.get_option_with_catch_all_fallback(config, username, 'oauth2_resource', fallback=None)
         oauth2_flow = AppConfig.get_option_with_catch_all_fallback(config, username, 'oauth2_flow')
         redirect_uri = AppConfig.get_option_with_catch_all_fallback(config, username, 'redirect_uri')
         redirect_listen_address = AppConfig.get_option_with_catch_all_fallback(config, username,
@@ -752,7 +753,7 @@ class OAuth2Helper:
 
         # because the proxy supports a wide range of OAuth 2.0 flows, in addition to the token_url we only mandate the
         # core parameters that are required by all methods: oauth2_scope and client_id
-        if not (token_url and oauth2_scope and client_id):
+        if not (token_url and client_id and (oauth2_scope or oauth2_resource)):
             Log.error('Proxy config file entry incomplete for account', username, '- aborting login')
             return (False, '%s: Incomplete config file entry found for account %s - please make sure all required '
                            'fields are added (at least token_url, oauth2_scope and client_id)' % (APP_NAME, username))
@@ -883,7 +884,8 @@ class OAuth2Helper:
                 auth_result = None
                 if permission_url:  # O365 CCG/ROPCG and Google service accounts skip authorisation; no permission_url
                     if oauth2_flow != 'device':  # the device flow is a poll-based method with asynchronous interaction
-                        oauth2_flow = 'authorization_code'
+                        if oauth2_flow != 'password':
+                            oauth2_flow = 'authorization_code'
 
                     permission_url = OAuth2Helper.construct_oauth2_permission_url(permission_url, redirect_uri,
                                                                                   client_id, oauth2_scope, username)
@@ -904,9 +906,10 @@ class OAuth2Helper:
                                    '`permission_url`' % (APP_NAME, username))
 
                 # note: get_oauth2_authorisation_tokens may be a blocking call (DAG flow retries until user code entry)
+                
                 response = OAuth2Helper.get_oauth2_authorisation_tokens(token_url, redirect_uri, client_id,
                                                                         client_secret, jwt_client_assertion,
-                                                                        auth_result, oauth2_scope, oauth2_flow,
+                                                                        auth_result, oauth2_scope, oauth2_resource, oauth2_flow,
                                                                         username, password)
 
                 if AppConfig.get_global('encrypt_client_secret_on_first_use', fallback=False):
@@ -1166,10 +1169,11 @@ class OAuth2Helper:
     @staticmethod
     # pylint: disable-next=too-many-positional-arguments
     def get_oauth2_authorisation_tokens(token_url, redirect_uri, client_id, client_secret, jwt_client_assertion,
-                                        authorisation_result, oauth2_scope, oauth2_flow, username, password):
+                                        authorisation_result, oauth2_scope, oauth2_resource, oauth2_flow, username, password):
         """Requests OAuth 2.0 access and refresh tokens from token_url using the given client_id, client_secret,
         authorisation_code and redirect_uri, returning a dict with 'access_token', 'expires_in', and 'refresh_token'
         on success, or throwing an exception on failure (e.g., HTTP 400)"""
+        
         if oauth2_flow == 'service_account':  # service accounts are slightly different, and are handled separately
             return OAuth2Helper.get_service_account_authorisation_token(client_id, client_secret, oauth2_scope,
                                                                         username)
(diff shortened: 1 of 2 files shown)


emailproxy.py
```diff
@@ -739,6 +739,7 @@ def get_oauth2_credentials(username, password, reload_remote_accounts=True):
         permission_url = AppConfig.get_option_with_catch_all_fallback(config, username, 'permission_url')
         token_url = AppConfig.get_option_with_catch_all_fallback(config, username, 'token_url')
         oauth2_scope = AppConfig.get_option_with_catch_all_fallback(config, username, 'oauth2_scope')
+        oauth2_resource = AppConfig.get_option_with_catch_all_fallback(config, username, 'oauth2_resource', fallback=None)
```
simonrob, line 742: Remove `, fallback=None` for consistency with retrieving the other catch-all values (the default fallback is `None`)

emailproxy.py
```diff
@@ -883,7 +884,8 @@ def get_oauth2_credentials(username, password, reload_remote_accounts=True):
                 auth_result = None
                 if permission_url:  # O365 CCG/ROPCG and Google service accounts skip authorisation; no permission_url
                     if oauth2_flow != 'device':  # the device flow is a poll-based method with asynchronous interaction
-                        oauth2_flow = 'authorization_code'
+                        if oauth2_flow != 'password':
+                            oauth2_flow = 'authorization_code'
```
simonrob, line 888: This isn't already present because the ROPCG flow doesn't have a `permission_url` value, so that is used as a way to select modes. Does the legacy version you're integrating with require a permission URL?

emailproxy.py
```diff
@@ -752,7 +753,7 @@ def get_oauth2_credentials(username, password, reload_remote_accounts=True):
 
         # because the proxy supports a wide range of OAuth 2.0 flows, in addition to the token_url we only mandate the
         # core parameters that are required by all methods: oauth2_scope and client_id
-        if not (token_url and oauth2_scope and client_id):
+        if not (token_url and client_id and (oauth2_scope or oauth2_resource)):
```
simonrob, line 756: Just to be clear, this variant of ROPCG uses a scope _or_ resource? Which email provider is this for, and is there any formal documentation you can point to?

emailproxy.py
```diff
@@ -904,9 +906,10 @@ def get_oauth2_credentials(username, password, reload_remote_accounts=True):
                                    '`permission_url`' % (APP_NAME, username))
 
                 # note: get_oauth2_authorisation_tokens may be a blocking call (DAG flow retries until user code entry)
+
```
simonrob, line 909: Remove

emailproxy.py
```diff
@@ -1192,28 +1196,45 @@ def get_oauth2_authorisation_tokens(token_url, redirect_uri, client_id, client_s
 
         if oauth2_flow != 'authorization_code':
             del params['code']  # CCG/ROPCG flows have no code, but we need the scope and (for ROPCG) username+password
-            params['scope'] = oauth2_scope
```
simonrob, line 1195: Restore this and instead delete if needed in the `password` flow, otherwise the CCG flow will not have a scope value.

emailproxy.py
```diff
@@ -1192,28 +1196,45 @@ def get_oauth2_authorisation_tokens(token_url, redirect_uri, client_id, client_s
...
                 params['device_code'] = authorisation_result['device_code']
                 expires_in = authorisation_result['expires_in']
                 authorisation_result['interval'] = authorisation_result.get('interval', 5)  # see RFC 8628, Section 3.2
             elif oauth2_flow == 'password':
                 params['username'] = username
                 params['password'] = password
+                if oauth2_resource:
+                    params['resource'] = oauth2_resource
+                else:
+                    params['scope'] = oauth2_scope
```
simonrob, line 1211: As noted above, instead do:

```python
if oauth2_resource:
    params['resource'] = oauth2_resource
    del params['scope']
```

emailproxy.py
```diff
@@ -1192,28 +1196,45 @@ def get_oauth2_authorisation_tokens(token_url, redirect_uri, client_id, client_s
 
         if oauth2_flow != 'authorization_code':
             del params['code']  # CCG/ROPCG flows have no code, but we need the scope and (for ROPCG) username+password
-            params['scope'] = oauth2_scope
             if oauth2_flow == 'device':
+                params['scope'] = oauth2_scope
```
simonrob, line 1200: Remove

emailproxy.py
```diff
@@ -1192,28 +1196,45 @@ def get_oauth2_authorisation_tokens(token_url, redirect_uri, client_id, client_s
...
+                # normalize expires_in to int (some endpoints return it as a string)
+                if 'expires_in' in token:
+                    try:
+                        token['expires_in'] = int(token['expires_in'])
+                    except (TypeError, ValueError):
+                        Log.error(
+                            'Invalid expires_in from token endpoint for account', username,':', token.get('expires_in')
+                        )
+                        token['expires_in'] = 0
+                return token
```
simonrob, line 1237: I agree that this is a potential issue (and it has cropped up before – e.g., #237), but I think I'd prefer to just cast the value when using it [as was done in response to that previous issue](https://github.com/simonrob/email-oauth2-proxy/blob/8570f520b011af9407e57edd17a6dc0200f0d657/emailproxy.py#L1292). I'll review and fix this separately as it isn't related to the rest of this PR. Best to revert this for now.

simonrob (commented):
Thanks for the PR! I've reviewed it and left some comments. Can you point to documentation for the syntax of this flow?

emailproxy.py
Zhehan-Z: This is for 21Vianet, a Chinese deployment variant of the original Microsoft 365 services (national cloud). OAuth using code is known to have compatibility issues with some clients due to a scope bug. In my tests the only way to make it work was through ROPCG with resources field. The server does not accept scope field in ROPCG flows. Documentation: https://www.cnblogs.com/CQman/p/16469223.html (in Chinese).

emailproxy.py
Zhehan-Z: This issue pops up in 21Vianet authentication in live testing. The original code would always return the Raw response without formatting on the first attempt, which led to the "int+chr" error.

Zhehan-Z:
> Thanks for the PR! I've reviewed it and left some comments. Can you point to documentation for the syntax of this flow?

The main documentation that I referenced to was this: https://www.cnblogs.com/CQman/p/16469223.html (in Chinese).

emailproxy.py
Zhehan-Z: After rechecking some documentation, I believe it does not. I'll revert this.

emailproxy.py
Zhehan-Z: It does not accept a scope value. A valid resource field is required.

Zhehan-Z:
I've updated the code to fix most of the problems mentioned in your comments. Thanks!

simonrob:
Thanks for the updates. I've fixed an indentation issue, cast `expires_in` to int and done some minor reformatting to match the code style. Please could you confirm that this version works with your provider?

Zhehan-Z:
It works! Checks have been completed on my own server, authentication  successful with a resource parameter sent to 21Vianet.

Zhehan-Z:
I have updated the configuration file with regards to this change.

simonrob:
Excellent – I made a few final very minor refinements, but I think this is now ready to go. Thanks for the contribution!

PR #359, opened 2025-05-31
wait until ssl handshake is completed to send data to server

jbraswell:
We had trouble getting email-oauth2-proxy to connect to the server when used with Istio. After a lot of debugging, we found that Istio delays opening a connection until it sees some data. It inspects that data to decide how to handle the traffic. This causes a timing issue: if the client sends data before the server's TLS handshake has occurred, Istio can't recognize the traffic correctly - it needs to see the TLS handshake first.

This fix stores any data sent before the handshake has occurred, then sends it all once the handshake is done. Open to other solutions.

### DIFF
--- a/emailproxy.py
+++ b/emailproxy.py
@@ -1364,6 +1364,11 @@ class SSLAsyncoreDispatcher(asyncore.dispatcher_with_send):
         self.ssl_handshake_errors = (ssl.SSLWantReadError, ssl.SSLWantWriteError,
                                      ssl.SSLEOFError, ssl.SSLZeroReturnError)
         self.ssl_connection, self.ssl_handshake_attempts, self.ssl_handshake_completed = self._reset()
+        self._ssl_not_ready_send_buffer = []
+
+    @property
+    def is_ssl_ready(self):
+        return True
 
     def _reset(self, is_ssl=False):
         self.ssl_connection = is_ssl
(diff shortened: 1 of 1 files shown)


simonrob:
Thanks for the PR. This isn't something I've seen myself, but I can see how it could cause an issue. Arguably this is more of a client problem, but I can also appreciate that it would be useful to work around in the proxy.

Maybe my own understanding isn't quite right, but I was under the impression that (except in cases where `STARTTLS` is used) no data is sent before the TLS handshake is initiated. So, are you referring to data sent after initiation and before the handshake completes?

If so, this sounds like a worthwhile fix, because I can see how that could lead to issues. Before merging, though, I'd like to find a way to replicate this sort of situation locally for testing. Do you have any suggestions?

jbraswell:
As for replicating locally, I think you'd probably have to run both a client and the proxy under kubernetes with istio running. Unfortunately I don't have a simple way for you to provision those pieces.

simonrob:
Ok – I'll think about how I could test this locally. In the meantime, if you do think of a way that would help with this that'd be very useful.

jbraswell:
It looks like we don't actually need these changes. We were able to completely workaround this issue by enabling tls on the proxy. The tls handshake between the client and the proxy forces the client to send some data, completing the connection through the istio sidecar. After that, everything seems to work just fine. It would be fine to close this PR.

simonrob:
Thanks for following up – I'm glad you were able to find a way to resolve this without the need to modify the proxy.

PR #384, opened 2025-10-08
get_oauth2_authorisation_code is expected to return two values

taskula:
get_oauth2_authorisation_code():
https://github.com/simonrob/email-oauth2-proxy/blob/9a27a5684ddc2ed1a2c6346fb38c381ee7e57d39/emailproxy.py#L1134-L1135

If device grant fails, we are also expected to return a boolean indicating a failure.

https://github.com/simonrob/email-oauth2-proxy/blob/9a27a5684ddc2ed1a2c6346fb38c381ee7e57d39/emailproxy.py#L909

Otherwise we end up with the following exception

"ValueError('too many values to unpack (expected 2)')"

### DIFF
--- a/emailproxy.py
+++ b/emailproxy.py
@@ -1132,7 +1132,7 @@ class OAuth2Helper:
             success, device_grant_result, permission_url, user_code = OAuth2Helper.start_device_authorisation_grant(
                 permission_url)
             if not success:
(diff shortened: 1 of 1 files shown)


simonrob:
Thanks for the fix!

Issue #411, opened 2026-03-10
Syslog identifier should be something other than "Email"


fangebee:
Hi, first let me thank you for this lifesaving project!
I've successfully deployed a headless server that allows old printers to continue functioning without trouble.

I've created a simple systemd unit file and it works fine too.

But, then, while trying to duplicate logs from journalctl to a dedicated plain log file in /var/log/emailproxy.log (we still love old stuff around here :) ) I struggled to craft the correct rsyslog configuration.

I was hoping I could use the following simple `/etc/rsyslog.d/emailproxy.conf` configuration file:

```
:programname, isequal, "emailproxy" -/var/log/emailproxy.log
& stop
```

But not luck :( 
Looks like rsyslog considers that the programname is "Email" because every log line starts with "Email OAuth 2.0 Proxy: ...". Therefore, the following configuration works:
```
:programname, isequal, "Email" -/var/log/emailproxy.log
& stop
```

It works, but it's not really satisfactory.

Maybe I'm doing something wrong, but for now I've patched the code to make sure the first word is "emailproxy".

PR #412, opened 2026-03-10
Make syslog identifier more intuitive

fangebee:
My fix for #411

### DIFF
--- a/emailproxy.py
+++ b/emailproxy.py
@@ -270,7 +270,7 @@ class Log:
     _LOGGER = None
     _HANDLER = None
     _DATE_FORMAT = '%Y-%m-%d %H:%M:%S:'
-    _SYSLOG_MESSAGE_FORMAT = '%s: %%(message)s' % APP_NAME
+    _SYSLOG_MESSAGE_FORMAT = '%s %%(message)s' % APP_SHORT_NAME
     _MACOS_USE_SYSLOG = False
 
     @staticmethod


emailproxy.py
```diff
@@ -270,7 +270,7 @@ class Log:
     _LOGGER = None
     _HANDLER = None
     _DATE_FORMAT = '%Y-%m-%d %H:%M:%S:'
-    _SYSLOG_MESSAGE_FORMAT = '%s: %%(message)s' % APP_NAME
```
simonrob, line 273: Changing the name makes sense, but why do we need to remove the `:` too?

fangebee: this woould make the `:` part of the syslog identifier, which is not desirable IMO

simonrob: Thinking about this further, would it be better to remove `_SYSLOG_MESSAGE_FORMAT` entirely, and just use `APP_SHORT_NAME` when [setting up the logger](https://github.com/simonrob/email-oauth2-proxy/blob/277643d28752f1fef2abbb0561b6ab140b994201/emailproxy.py#L278)? That would fall back to the logger's [standard format](https://github.com/python/cpython/blob/a778fd600445f78a6f3f9bc97d44a9096a220ff0/Lib/logging/__init__.py#L546), which is likely to be more widely expected/supported.

simonrob: Just wondering whether you had any thoughts on this?

fangebee: sorry for not replying, I need some time to prepare a staging platform for testing things, because the platform where I deployed emailproxy is now in production.

PR #413, opened 2026-04-08
Add Thunderbird Android and K-9 mail for client details

AdityaGarg8:


simonrob:
Thanks for the contribution – interesting that these use different tokens to the main Thunderbird app. I removed the older Thunderbird one at the same time as that is likely to be expired at some point.

AdityaGarg8:
> Thanks for the contribution – interesting that these use different tokens to the main Thunderbird app. I removed the older Thunderbird one at the same time as that is likely to be expired at some point.

Mobile device applications are required to have a seperate client id and secret for the purpose of getting a redirect uri that dumps the code to the app itself.

simonrob:
Most providers allow you to add multiple redirect URIs, though, and you specify the redirect in the token request, so it wouldn't normally be a problem to use the same client ID. Maybe some providers don't support this, so they decided to separate them all out.

AdityaGarg8:
> Most providers allow you to add multiple redirect URIs, though, and you specify the redirect in the token request, so it wouldn't normally be a problem to use the same client ID. Maybe some providers don't support this, so they decided to separate them all out.

It's like on Android google requires you to also have an android signature linked to you app. Mobile devices are just a mess with different providers having different rules.

PR #424, opened 2026-08-27
feat: add --no-tray option using GTK window interface on Linux

iacobucci:
### Summary

In certain Linux desktop environments (such as GNOME 50+ with AppIndicator/StatusNotifierItem), system tray icons can be buggy under some conditions. This PR introduces an
optional `--no-tray` CLI flag that displays a standalone GTK window interface instead of the tray icon on Linux for configuring emailproxy.

### Key Changes

- **CLI Flag `--no-tray`**: Added a new `--no-tray` argument under GUI appearance options.
- **Default Behavior Retained**: The proxy continues to use the system tray icon (`pystray`) by default.
- **GTK Window Interface (`GtkProxyWindow`)**: Implemented a native GTK 3 interface using PyGObject when `--no-tray` is enabled on Linux:
	- Displays proxy server status (IMAP/POP/SMTP) and listening addresses.
	- Shows configured accounts and last authenticated activity timestamps.
	- Manages pending account authorisation requests with interactive buttons.
	- Provides toggles for external authorisation mode, debug mode, and start at login.
- **Non-blocking Window Close**: Connected `delete-event` to `hide_on_delete()` so closing the GTK window hides it while keeping background proxy threads running. Process exit is
managed via the "Quit Email OAuth2 Proxy" button or signals.
- **Dependencies & Documentation**: Added `PyGObject` for Linux GUI requirements in `requirements-gui.txt` and documented `--no-tray` in `README.md`.

### Testing

- Verified `--help` and default tray behavior on startup.
- Tested `--no-tray` startup, UI interactions (refresh activity, reload config, authorisation prompts, toggles), hiding on window close, and clean exit via the Quit button.


simonrob:
Thanks for the contribution. I'm open to UI improvements, but this one feels far too focused on a single platform, and also brings with it more maintenance burden going forwards (for example, the need to be more careful when using `self.icon`).

As a result, I've decided not to accept this change. Anyone needing this feature could of course just use the version in this PR, or simply generate their own solution.

